PR #26, opened 2014-07-07
analytics.page documentation / library mismatch

skaapgif:
From https://segment.io/docs/libraries/node/: 

> A few properties are automatically added to each .page() call. For example, if you call:
>
> analytics.page('Pricing');
> We will translate that to the following without any extra work from you:
>
> analytics.track('Pricing', {
>   title: 'Segment.io Pricing',
>   url: 'https://segment.io/pricing',
>   path: '/pricing',
>   referrer: 'https://segment.io'
> });

I was sceptic (but hopeful) when I saw .page() promised to do all this without being passed a request object or headers. It looks like all the documentation for page is from the browser library.

Having said that, it would be very handy for the library to extract these from a headers argument though, any plans for adding this? I'm having to write this for a wrapper around analytics-node anyway, so won't mind creating a PR to integrate it into the library.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -11,6 +11,62 @@ var version = require('../package.json').version;
 global.setImmediate = global.setImmediate || process.nextTick.bind(process);
 
 /**
+ * Internal helper functions for creating `pure` functions
+ */
+_clone = function(source) {
+  var dest = {};
+  for (key in source) {
+    dest[key] = source[key];
+  }
+  return dest;
+}
+
+_defaults = function(source, defaults) {
+  source = _clone(source);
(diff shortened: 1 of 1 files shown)


sperand-io:
Hey @skaapgif thanks for the reference/prototype implementation here — this is super interesting and something I've added to our internal tracking system to consider how we might best implement it. Another option would be for us to start publishing some extensible middlewares for common frameworks. In any case, really appreciate this and will be sure to ping you if we decide to move forward in this direction!

skaapgif:
Great, I'm glad it's been helpful.

PR #33, opened 2014-10-06
add retries

travisjeffery:
we should be able to just toss in superagent-retry: https://github.com/segmentio/superagent-retry

any dupes should be handled via our API de-duping


stephenmathieson:
@calvinfo going to close for now, since this is multiple years old, very conflicted and we already retry.

PR #39, opened 2015-02-05
Add installation instructions to readme

Nijikokun:
This is frustrating as a user to find, you have to go to the repo, go to the documentation, go to getting started inside of that documentation, and finally you will find this line.


gjohnson:
Thanks, good call!

PR #40, opened 2015-02-18
add browserify, without proxy working yet

ianstormtaylor:
Cc @travisjeffery trying to solve this: https://trello.com/c/bN7m6CNd/407-browserify-analytics-node

But for some reason the `superagent-proxy` library is erroring out when it's included in the browserify build. I think it has to do with it using `regenerator` under the hood unfortunately, but can't seem to get that working with browserify transforms. Could you take a look?

Other than that, this is good to go for anyone using Chrome Extensions, Cordova, etc.


lib/index.js
```diff
@@ -3,7 +3,7 @@ var clone = require('clone');
 var debug = require('debug')('analytics-node');
 var noop = function(){};
 var request = require('superagent');
-require('superagent-proxy')(request);
+// require('superagent-proxy')(request);
```
ianstormtaylor: Here's the part where commenting out this line means that the browserify bundled doesn't error out on `fs.readFileSync` not existing.

test/index.js
```diff
@@ -1,10 +1,362 @@
...
+    //     flushAfter: Infinity,
+    //     proxy: 'http://localhost:4064'
+    //   });
+    //   a.enqueue('type', { event: 'test' }, noop);
+    //   a.flush(function(err, data){
+    //     // our proxy turns all responses into 408/errs
+    //     assert(err);
+    //     done();
+    //   });
+    // });
```
ianstormtaylor: Also had to comment out this test for now, but should be uncommented when we get the proxying to play nice.

travisjeffery:
Awesome! We're good to go now!

PR #42, opened 2015-03-13
Corporate Proxy Support

kauegimenes:
I included the module global-tunnel to provide proxy support using the variable HTTP_PROXY.

Use globa-tunnel to provide proxy support.
The http_proxy environment variable will be used if the first parameter to globalTunnel.initialize is null.


sperand-io:
Thanks! I've marked this for consideration in our internal tracking system and will be sure to ping you if we add support. FWIW — analytics-node.js is a browserified build of the library, but any changes we'd make would need to be added inside the `/lib` directory.

PR #47, opened 2015-09-14
remove the duplicated keyword

yorkie:
/cc @travisjeffery

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,7 @@
     "segment.io",
     "segmentio",
     "client",
-    "driver",
-    "analytics"
+    "driver"
   ],
   "main": "lib/index.js",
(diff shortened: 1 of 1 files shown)


travisjeffery:
Thanks @yorkie!

yorkie:
BTW, a question about the next of this project, actually I was using segment.io, but with a un-expectable reason, have to tend to use another analytics service from China mainland, plus I found you have a port to set the endpoint's host/name, so do you want a pr to abstract the different ports between most compatible services?

For instance, I'm working on https://github.com/weflex/zhuge-node which is a fork from this repo. Ideally I hope there is 1 standard/core module that every/most analytics services should follow :)

PR #49, opened 2015-11-23
Found known vulnerabilities in dependencies

jesusprubio:
Hi, I'm using "analytics-node" in a project and I realized that this repo is using libraries with known vulnerabilities. So I've added [Node Security Project](https://nodesecurity.io) support and I've also updated the conflictive ones.

Could you please take a look? :)

captura de pantalla 2015-11-23 a las 9 37 05

Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Binaries.
-browserify = ./node_modules/.bin/browserify
-mocha = ./node_modules/.bin/mocha
+browserify = ./node_modules/browserify/bin/cmd.js
+mocha = ./node_modules/mocha/bin/mocha
 
 # Build the browserify bundle.
 analytics-node.js: node_modules lib/index.js
--- a/Makefile
+++ b/Makefile
@@ -22,5 +22,8 @@ test: node_modules
 clean:
 	@rm analytics-node.js
 
+nsp:
+	@$(nsp) check
+
 # Phonies.
 .PHONY: test
--- a/package.json
+++ b/package.json
@@ -29,14 +29,15 @@
   },
   "devDependencies": {
     "async": "~0.9.0",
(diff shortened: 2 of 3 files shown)


Nijikokun:
:+1: Super urgent, should be merged, can't believe it hasn't yet.

calvinfo:
Hey @Nijikokun thanks for the PR! I'm running through it now and looking at all the tests.

Out of curiosity, where you able to find any way to exploit it? As far as I can tell, the `ms` module is only ever being called with numbers by the `debug` module and isn't ever parsing a regex. I couldn't see any sort of DOS from the way the code currently operates–curious to hear what you found.

Makefile
```diff
@@ -1,6 +1,6 @@
 # Binaries.
-browserify = ./node_modules/.bin/browserify
-mocha = ./node_modules/.bin/mocha
+browserify = ./node_modules/browserify/bin/cmd.js
+mocha = ./node_modules/mocha/bin/mocha
```
calvinfo: Did these not work as-is? I'd rather keep them in the `/.bin` folder in case the packages ever decide to change the filename of their binary

calvinfo:
Everything looks good to me otherwise! Down to merge after that.

Makefile
Nijikokun: This should be changed back to the `/.bin` folder

Nijikokun:
If you merge this, I can submit a pull request to change the paths back.

calvinfo:
Thanks, merged and updated!

Makefile
jesusprubio: Sure, my fault :)

jesusprubio:
:+1:

PR #54, opened 2016-01-22
Reset flushing timer only during flush

### LINKED ISSUE #53: Flush isn't behaving according to the docs
According to the documentation:

```
By default, our library will flush:

The very first time it gets a message.
Every 20 messages (controlled by options.flushAt).
If 10 seconds has passed since the last flush (controlled by options.flushAfter)
```

However the actual implementation resets the timeout every single time an event is enqueued:

https://github.com/segmentio/analytics-node/blob/master/analytics-node.js#L183-L185

What this means is that if you have `flushAt` set to 20, and `flushAfter` set to 10 seconds then each time a message arrives and is enqueued the timeout that controls `flushAfter` is reset to 10 seconds again.

So:

1) A message arrives in queue, and the timer is set for 10 seconds
2) 9 seconds pass, and another message arrives in queue. The timer is reset for another 10 seconds in the future, now 19 seconds since the last flush.
3) 9 seconds pass, another event arrives, and the flush timer is reset again, now triggering at 28 seconds since the last flush.

So with the default settings the queue can be stored for up to 200 seconds before flushing rather than the desired 10 seconds.

The correct behavior would be to set the timer on the first enqueued item, and then leave it as is, so that the next flush happens exactly 10 seconds later. If I have time I'll open a PR to fix this.

nathanpeck:
Makes the behavior of the `flushAfter` setting adhere to the documentation to fix #53

```
By default, our library will flush:

The very first time it gets a message.
Every 20 messages (controlled by options.flushAt).
If 10 seconds has passed since the last flush (controlled by options.flushAfter)
```

### DIFF
--- a/analytics-node.js
+++ b/analytics-node.js
@@ -129,7 +129,10 @@ Analytics.prototype.alias = function(message, fn){
 
 Analytics.prototype.flush = function(fn){
   fn = fn || noop;
-  if (this.timer) clearTimeout(this.timer);
+  if (this.timer) {
+    clearTimeout(this.timer);
+    this.timer = null;
+  }
   if (!this.queue.length) return setImmediate(fn);
 
   var items = this.queue.splice(0, this.flushAt);
--- a/analytics-node.js
+++ b/analytics-node.js
@@ -129,6 +129,7 @@ Analytics.prototype.alias = function(message, fn){
 
 Analytics.prototype.flush = function(fn){
   fn = fn || noop;
+  if (this.timer) clearTimeout(this.timer);
   if (!this.queue.length) return setImmediate(fn);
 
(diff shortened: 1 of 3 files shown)


sperand-io:
LGTM — @ndhoule mind giving another set of eyes?

wilzh40:
Is it possible to merge this? Our CLI application hangs when it is supposed to quit...calling flush() does not flush the queue immediately.

vadimdemedes:
Hey @nathanpeck, would you be able to update your PR to resolve the merge conflict?

vadimdemedes:
@nathanpeck No worries, I rebased the PR for you and opened a new one - https://github.com/segmentio/analytics-node/pull/103. Thanks for your detailed issue report and work to fix it!

PR #56, opened 2016-03-02
increase messageId entropy, and include one id per message

calvinfo:
This changes the behavior of setting end-to-end messageIds, so
that the client library now generates individual ones per message
rather than a single group of them per batch.

It also increases the entropy, increasing them to 32 characters.


vinceprignano:
LGTM

PR #57, opened 2016-03-03
Adding a CLI

dominicbarnes:
While working on a revamped simulator, I've decide it's best to create docker containers for each library. In addition, it's way easier `docker run` things that have a CLI, so I've added one here. It also turned out this tool could be useful for success engineers or for debugging, so I've created this as a separate PR after discussing with @calvinfo.

The big idea is that you add a `$SEGMENT_WRITE_KEY` to your env and then you can run the following commands:

```bash
$ analytics track <event> [options]
$ analytics identify [options]
$ analytics group [options]
$ analytics alias [options]
$ analytics page [options]
```

All the options are consistent throughout the commands, which you can see by using `--help`.

**NOTE:** This adds 1 more dependency to `analytics-node` itself, which is [commander](https://www.npmjs.com/package/commander). (it's a lightweight) This does not inflate the browserify build at all. (since it's only being used in the `bin/` dir)

### DIFF
--- a/bin/analytics
+++ b/bin/analytics
@@ -1,0 +1,129 @@
+#!/usr/bin/env node
+
+var Analytics = require('..');
+var assert = require('assert');
+var pkg = require('../package.json');
+var program = require('commander');
+
+program
+  .version(pkg.version)
+  .option('-w, --write-key <key>', 'the segment write key to use');
+
+program
+  .command('track <event>')
+  .description('track a user event')
+  .option('-u, --user <id>', 'the user id to send the event as')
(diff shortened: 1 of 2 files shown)


stephenmathieson:
should we support `--write-key=ABCDEFG`?

stephenmathieson:
i think it'd be cool to use commander's `.command()`s.  this would let us remove a bunch of the copypasta in `./bin/*` too

dominicbarnes:
@stephenmathieson both suggestions have been applied :)

stevenmiller888:
This is cool. Still planning to merge?

PR #64, opened 2016-07-29
Remove unused `superagent-proxy` dependency

stwiname:


neiker:
Proxy remaining code also should be removed from test/server.js

stephenmathieson:
#89

PR #68, opened 2016-09-08
improving flush, adding onAllDataFlushed, removing a global

adrian-gierakowski:
1. flush : when calling twice one after another ( or when flush called
   manually directly follows a flush triggered internally ), the second
   flush would trigger its callback before the previous could finish
   sending its data. This is misleading since you would usually expect
   data data you have just enqueued, has been safely uploaded once the
   flush callback is called. The commit improves this behaviour by waiting
   for currently pending flush to finish before executing the callback of
   the consecutive flush ( as long as there was no events in the queue when
   the second flush was called ).
2. onAllDataFlushed : even with the above improvement, if we rally want
   to be sure that there is no pending data ( for example if we want to
   close the app ), we need to do a little bit more work, since even though
   a flush is done, it doesn’t mean that new data has not been queued and
   flushed in the meantime ( between a call to flush and it callback ).
   onAllDataFlushed registers a callback ( listener ) which fires once,
   whenever a flush is done and there are no more messages in the queue.
3. replacing setImmediate global with local var : there is no reason to set a global in order to use this function within the lib. Its not nice for an npm module to do that

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -34,6 +34,10 @@ function Analytics(writeKey, options){
   assert(writeKey, 'You must pass your Segment project\'s write key.');
   options = options || {};
   this.queue = [];
+  this.flushesInProgressCount = 0;
+  this.flushesInProgress = {};
+  this.mostRecentFlushId = null;
+  this.allDataFlushedListeners = [];
   this.writeKey = writeKey;
   this.host = options.host || 'https://api.segment.io';
   this.flushAt = Math.max(options.flushAt, 1) || 20;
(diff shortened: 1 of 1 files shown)


adrian-gierakowski:
all those changes are only in the node implementation, I can copy them over to the browser file
I can also add doc strings for the private methods if needed

lib/index.js
```diff
@@ -10,7 +10,7 @@ var uid = require('crypto-token');
 var version = require('../package.json').version;
 var extend = require('lodash').extend;
 
-global.setImmediate = global.setImmediate || process.nextTick.bind(process);
```
timelf123: This should be the default 👍

adrian-gierakowski:
@timelf123 is there anything you'd like me to do before you can merge these changes? thanks!

timelf123:
@adrian-gierakowski I'm not employed by segment, just giving my two cents for what it's worth

f2prateek:
@adrian-gierakowski can you send these as seperate PRs?  I'd like to discuss these changes seperately:

1. flush : I think this is reasonable to accept as implemented in the PR.
2. onAllDataFlushed: I'd like to think through the API design of this a bit more.
3. replacing setImmediate global with local var: I think this unintentionally might be a breaking change?

PR #77, opened 2017-01-25
Bump up lodash version so it helps other projects with conflicting versions

filipesperandio:


timelf123:
was set to `~4.17.2` in latest master

PR #78, opened 2017-02-02
Replace validation with `@segment/loosely-validate-event`

stephenmathieson:
I needed this validation logic elsewhere, so I pulled it out and moved it to the [loosely-validate-event](https://github.com/segmentio/loosely-validate-event) module. The code is basically a 1:1, just stylistic changes. This patch pulls in this new module and offloads all validation logic to it.

@segmentio/gateway

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -4,11 +4,10 @@ var debug = require('debug')('analytics-node');
 var noop = function(){};
 var request = require('superagent');
 require('superagent-retry')(request);
-var type = require('component-type');
-var join = require('join-component');
 var uid = require('crypto-token');
 var version = require('../package.json').version;
 var extend = require('lodash').extend;
+var validate = require('@segment/loosely-validate-event');
 
 global.setImmediate = global.setImmediate || process.nextTick.bind(process);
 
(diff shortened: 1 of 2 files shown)


f2prateek:
LGTM

Could we not refactor and reuse https://github.com/segmentio/api-validator?

stephenmathieson:
> Could we not refactor and reuse https://github.com/segmentio/api-validator?

We _could_, but it's private (I assume for a reason?). In addition, that thing requires access to a database. O_o

f2prateek:
yeah it's private because of db-access. I think if we split the db stuff out it could be useful to just re-use that.

stephenmathieson:
That's fair. I'd rather not have to mess with updating the tracking API to get this change made tho. Are you cool with this landing as-is, @f2prateek?

f2prateek:
Yup good for now!

distracteddev:
@stephenmathieson Seems like we can't npm install this newly created package. Perhaps it was not published correctly? I can see the package here:

https://www.npmjs.com/package/@segment/loosely-validate-event

But running `npm info @segment/loosely-validate-event` returns an "Invalid package error".

PR #81, opened 2017-02-08
remove trailing slash on `.host`

stephenmathieson:
I just spent over an hour trying to figure out why an event wasn't showing up. The reason was I was POSTing to `https://api.segment.tld//v1/batch` rather than `https://api.segment.tld/v1/batch`.

@segmentio/gateway @fouad


f2prateek:
LGTM. Might be worth adding a test case.

stephenmathieson:
@f2prateek I agree 100%. Will do in a sec.

f2prateek:
:shipit:

PR #84, opened 2017-03-11
Add node version to messages via _metadata

### LINKED ISSUE #79: Include the Node version in the event somehow
Would be nice for internal metrics to see what versions of node our customers are actually using

Dorian:
FIx #79 :)

I first tried to add it to the context, as it's where the library version is, but per #79 it should be in `_metadata`.

CC: @myclamm 

(mostly based on this github search: https://github.com/search?q=org%3Asegmentio+_metadata&type=Code)

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -10,8 +10,7 @@ var id = 'id';
 var context = {
   library: {
     name: 'analytics-node',
-    version: require('../package.json').version,
-    nodeVersion: process.versions.node
+    version: require('../package.json').version
   }
 };
 
@@ -128,8 +127,7 @@ describe('Analytics', function(){
       assert.deepEqual(a.queue[0].message.context, {
         library: {
           name:'analytics-node',
-          version: require('../package.json').version,
-          nodeVersion: process.versions.node
+          version: require('../package.json').version
         },
         name: 'travis'
       });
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -6,10 +6,11 @@ var server = require('./server');
 var a;
 var noop = function(){};
 var id = 'id';
+
 var context = {
   library: {
     name: 'analytics-node',
-    version: require('../package.json').version
+    version: require('../package.json').version,
```
f2prateek: Unnecessary comma?

test/index.js
```diff
@@ -128,7 +129,7 @@ describe('Analytics', function(){
           name:'analytics-node',
           version: require('../package.json').version
         },
-        name: 'travis'
+        name: 'travis',
```
f2prateek: Unnecessary comma?

Dorian:
@f2prateek It would be nice to have an eslint config in your projects BTW (this is what I got when inspecting the non-generated JS files: https://gist.github.com/Dorian/7aa3f19be73d6826ba9af9e549cf4757 (e.g. `bin/analytics` `lib/index.js` `test/index.js` `test/server.js`))

stephenmathieson:
>  It would be nice to have an eslint config in your projects BTW 

I totally agree here. We've been migrating a lot of this stuff to use `standard`, so that'll probably happen here soon-ish. Thanks for the config tho!!

PR #90, opened 2017-03-14
add `mock.js`

### LINKED ISSUE #75: Provide a test mode or a mock class
Out-of-the-box, there is no supported way to test against Segment's `analytics-node` module without a connection to a Segment server.  In your own tests, you guys startup a server process that listens on localhost to test against.  This isn't a practical solution for consumers, and it's a better practice to provide a mock shim to test against.  Most other libraries support this, please add this functionality.

I think I can work around this by monkey-patching `flush`:
```
analytics.flush = function(fn) {
  process.nextTick(() => { fn(null, []); });
};
```

This is pretty lame though, a supported mock stub would be greatly appreciated.  Thanks.

stephenmathieson:
This patch adds a `mock.js` file, which we'll be able to use in
development/test/etc (anywhere you don't want to send real events).

I'm envisioning usage looking something like:

```js
const { NODE_ENV = 'development' } = process.env

const Analytics = NODE_ENV === 'production'
  ? require('analytics-node')
  : require('analytics-node/mock')
const analytics = new Analytics('abc write key', { some: 'options' })

analytics.track({ userId: 'aaa', event: 'Did Something' })
```

@segmentio/gateway @thesmart 

Closes #75.

### DIFF
--- a/mock.js
+++ b/mock.js
@@ -1,0 +1,25 @@
+var Analytics = require('./')
+
+var temp = new Analytics('fakekey')
+
+module.exports = AnalyticsMock
+
+function AnalyticsMock () {
+  if (!(this instanceof AnalyticsMock)) {
+    return new AnalyticsMock()
+  }
+}
+
+for (var key in temp) {
+  var fn = temp[key]
+  if (typeof fn === 'function') {
+    AnalyticsMock.prototype[key] = mock
+  }
+}
+
+function mock () {
+  var callback = arguments[arguments.length - 1]
+  if (typeof callback === 'function') {
+    process.nextTick(callback)
+  }
+}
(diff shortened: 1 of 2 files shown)


f2prateek:
I can't tell, but what's the behaviour if you call `analyticsMock.track({})`. Does it throw an exception like the real client or silently swallow it?

stephenmathieson:
The "real" client would throw. I think you're on to something here - this could cause all sorts of confusion if things worked in dev/test but threw in prod. I'll refactor tomorrow after I get some sprint work done.

stephenmathieson:
Thinking about this a bit further, I don't think it's worth us supporting a mock. The monkey patch from #75 is only 3 SLOC - users can add it on their own.

f2prateek:
so can we close #75 as well?

PR #104, opened 2017-07-24
Remove browserify build

vadimdemedes:
Since @stevenmiller888 and me couldn't find a source problem why it was added in the first place, we decided to remove it. There's a chance it wasn't even related to analytics-node or the problem the users were experiencing was fixed since then.

If the issue appears again, we'll have fresh reports to work on.


stephenmathieson:
👍

PR #110, opened 2017-07-25
Convert to ES6

vadimdemedes:
This PR brings a breaking change - inability to create an instance without a `new` keyword, which is a JavaScript limitation.

Also removes the dependency on `lodash/extend` and `clone`, since those can be replaced with `Object.assign()`.

Tests are failing, because I haven't modified them for the above mentioned change. I would like to merge #108 first, to avoid a merge conflict.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,205 +1,15 @@
-var assert = require('assert')
-var clone = require('clone')
-var debug = require('debug')('analytics-node')
-var noop = function () {}
-var request = require('superagent')
-var uid = require('crypto-token')
-var version = require('../package.json').version
-var extend = require('lodash/extend')
-var validate = require('@segment/loosely-validate-event')
-var removeSlash = require('remove-trailing-slash')
+'use strict'
 
-global.setImmediate = global.setImmediate || process.nextTick.bind(process)
+const assert = require('assert')
+const removeSlash = require('remove-trailing-slash')
+const validate = require('@segment/loosely-validate-event')
+const request = require('superagent')
+const debug = require('debug')('analytics-node')
+const uid = require('crypto-token')
+const version = require('../package').version
 
-/**
- * Expose an `Analytics` client.
- */
-
-module.exports = Analytics
-
-/**
- * Initialize a new `Analytics` with your Segment project's `writeKey` and an
- * optional dictionary of `options`.
- *
- * @param {String} writeKey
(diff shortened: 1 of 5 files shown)


lib/index.js
```diff
@@ -1,216 +1,234 @@
...
 }
 
-/**
- * Send a group `message`.
- *
- * @param {Object} message
- * @param {Function} [fn] (optional)
- * @return {Analytics}
- */
+class Analytics {
```
stephenmathieson: using a `class` here is a breaking change, since you're now _required_ to use `new`:

```js
class Analytics {}
const a = Analytics()
// => TypeError: Class constructor Analytics cannot be invoked without 'new'
```

that said, I'm 100% for this, but we'll need to bump the major version and (probably) let the success team know about the change

lib/index.js
```diff
@@ -1,216 +1,234 @@
...
+   *
+   * @param {String} writeKey
+   * @param {Object} [options] (optional)
+   *   @property {Number} flushAt (default: 20)
+   *   @property {Number} flushAfter (default: 10000)
+   *   @property {String} host (default: 'https://api.segment.io')
+   */
+
+  constructor (writeKey, options) {
+    options = options || {}
```
stephenmathieson: could use a default param here (`constructor (writeKey, options = {})`) ;]

lib/index.js
stevenmiller888: Yup, major version is a given. Will let Success know.

lib/index.js
vadimdemedes: Unfortunately we can't, because default function params aren't supported in Node.js v4.

lib/index.js
vadimdemedes: The next release would be a major anyway, since we no longer include browserify build and also because of https://github.com/segmentio/analytics-node/pull/105, which can also break someone's code.

vadimdemedes:
PR is ready to go!

lib/index.js
stephenmathieson: do enough folks still use node 4 for that to matter?

nbd either way. this is awesome- keep it up dude!

vadimdemedes: Yeah, there are still quite a few people using it. Supporting it is no hassle for us at this point, we have to sacrifice only default params, so I think it should be fine. In April 2018 when v4 is discontinued, we'll drop it.

Issue #95, opened 2017-05-10
`flush()` callback not reliable if there's a pending flush, can lead to out-of-order events


bradvogel:
Consider the following code:

```
var Analytics = require('analytics-node');
var analytics = new Analytics('<write key>', {
  flushAt: 1
});

analytics.identify({
  userId: '019mr8mf4r',
  traits: {
    name: 'Michael Bolton',
    email: 'mbolton@initech.com',
    plan: 'Enterprise',
    friends: 42
  }
}, function() {
  console.log('done with identify');
});

analytics.flush(function() {
  console.log('done flushing')
});
```

I would expect `done flushing` to be logged _after_ `done with identify`, but it is not. Intuitively I'd expect `flush()` to only call its callback once _all_ pending calls have been flushed.

**Why is this an issue?**

Because we need to be certain that the `identify()` call completes _before_ calling `track()`. Otherwise the `track()` won't carry the trait information sent in `identify()`.

So, for instance, the following code:

```
var Analytics = require('analytics-node');
var analytics = new Analytics('<write key>', {
  flushAt: 1
});

analytics.identify({
  userId: '019mr8mf4r',
  traits: {
    name: 'Michael Bolton',
    email: 'mbolton@initech.com',
    plan: 'Enterprise',
    friends: 42
  }
}, function() {
  console.log('done with identify');
});

analytics.flush(function() {
  analytics.track({
    userId: '019mr8mf4r',
    event: 'first event after identify'
  }, function() {
    console.log('done with first event')
  });
});
```

will log:
```
sending event that I want to go after identify()
done with first event
done with identify
```

This is wrong because the `track()` call is completing _before_ the `identify()` completes, despite the code flushing in between the calls.

I understand that the fix is to call `track()` in the callback of `identify()`, and the call `flush()` after without a callback. But this isn't intuitive at all to the user.

vadimdemedes:
Since you set `flushAt` to `1`, there's no need to call `flush()` manually, it will be called for you after each `identify()`, `track()` and other tracking methods.

As you correctly pointed out, in this specific case, calling `track()` in the callback of `identify()` would fix this.

> Because we need to be certain that the `identify()` call completes _before_ calling `track()`. Otherwise the `track()` won't carry the trait information sent in `identify()`.

[...]

PR #114, opened 2017-07-26
Flush after identify or group

vadimdemedes:
Fixes https://github.com/segmentio/analytics-node/issues/95.

As discussed, flushing after `identify()` is necessary in order to preserve traits for future messages. When looking through the docs, I saw that `group()` also has traits, so I added it too.

One question though, should we also flush after `alias()`?

### DIFF
--- a/index.js
+++ b/index.js
@@ -182,6 +182,11 @@ class Analytics {
       return
     }
 
+    if (type === 'identify' || type === 'group') {
+      this.flush()
+      return
+    }
+
     if (this.queue.length >= this.flushAt) {
       this.flush()
     }
--- a/test.js
+++ b/test.js
@@ -143,6 +143,23 @@ test('enqueue - flush on first message', t => {
   t.true(client.flush.calledTwice)
 })
 
+test('enqueue - flush on identify or group message', t => {
+  const client = createClient({ flushAt: 10 })
+  spy(client, 'flush')
+
+  client.enqueue('identify', {})
+  t.true(client.flush.calledOnce)
+
+  client.enqueue('group', {})
+  t.true(client.flush.calledTwice)
+
+  client.enqueue('track', {})
+  client.enqueue('page', {})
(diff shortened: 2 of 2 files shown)


stevenmiller888:
The end-tools stitch together the anonymous and identified histories irrespective of when the connection is established. Also, we don't guarantee order in the pipeline anyway so I don't think it makes sense for the libraries to handle this.

stevenmiller888:
I can see why this is unexpected for the users of this library though.

vadimdemedes:
In that case, the problem mentioned in #95 shouldn't occur?

> Because we need to be certain that the identify() call completes before calling track(). Otherwise the track() won't carry the trait information sent in identify().

stevenmiller888:
Correct!

stevenmiller888:
It could occur in that events may be delivered out of order. But the end tool should be stitching together the histories so it's okay. Probably worth asking a follow up on why it's important to him that events are delivered in order.

vadimdemedes:
I think it's important for him, because of this:

> Otherwise the track() won't carry the trait information sent in identify().

I'll follow up and confirm that this issue actually happens. Maybe it's a bug on the backend, if it didn't stitch the history correctly.

stephenmathieson:
TLDR; this will not work as expected 

as stated previously, there is no guarantee that events will be processed in the same order the arrive at Segment's "front door" (the tracking API: api.segment.io). there are tons of things that could change the order events are processed. to list a few: an event could get rejected then retried, an NSQ node could explode, a worker could OOM and stop accepting messages while it waits for others to take some load...

because of this, adding workarounds/false promises in the client libraries seems like the wrong move. instead we ought to explain this better in our documentation and make sure success is aware.

vadimdemedes:
Sounds reasonable, thanks @stephenmathieson and @stevenmiller888 for explaining this!

PR #117, opened 2017-07-28
Add changelog back

vadimdemedes:
This reverts commit 547a001e4a5f72bd94fb06155f37f5ee695155da.

@f2prateek what tool do you use to generate a changelog? I searched for `git-changelog`, but haven't found one that generates a similar output to what we have.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,0 +1,108 @@
+
+2.4.1 / 2017-05-05
+==================
+
+  * Preventing webpack to bundle all lodash lib (#94)
+  * README: fix header
+  * README: fix badge
+  * package: upgrade dependencies (#89)
+  * standard (#88)
+  * yarn (#85)
+  * analytics-node.js@2.4.0 [ci skip]
+
+2.4.0 / 2017-03-13
+==================
+
+  * test: add `_metadata` to "screen" assertion (#87)
+  * replace Travis with CircleCI (#86)
+  * add support for screen events (#82)
+  * use correct optional syntax for JSDOC (#83)
+
+2.3.0 / 2017-03-13
+==================
+
+  * Add node version to messages via _metadata (#84)
+  * remove trailing slash on `.host` (#81)
+
+2.2.0 / 2017-02-06
+==================
+
+  * Replace validation with `@segment/loosely-validate-event`
+  * add releasing notes
+
+2.1.1 / 2017-01-31
+==================
+
+  * Merge pull request #74 from alexstrat/master
+  * Update dependencies
+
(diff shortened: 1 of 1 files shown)


f2prateek:
For https://github.com/segmentio/analytics-android/blob/master/CHANGELOG.md, I generate the scafold with git-changelog, but then edit it by hand.

vadimdemedes:
I mean where can I download `git-changelog`, since it's not a built in command?

vadimdemedes:
Cool, thanks!

History.md
```diff
@@ -0,0 +1,133 @@
+
+v3.0.0 / 2017-07-27
+===================
+
+  * Flush on first message (#113)
```
f2prateek: Can we more diligent about the changelog here? My workflow is to typically start with git-changelog, and cleaning it up. e.g. here this could be (take my descriptions with a grain of salt, I'm not a js dev):

```
Version 3.0.0 (July 28th, 2017)
==============================
 * [New](https://github.com/segmentio/analytics-node/pull/110): Migrate to ES6. This is a breaking change. The `new keyword` is now required while initializing the client.
```
// Previously:
const a = Analytics()

// Now:
const a = new Analytics()
```

 * [Fix](https://github.com/segmentio/analytics-android/pull/112): Rename `flushAfter` setting to `flushInterval`. This is a breaking change.
 * [Fix](https://github.com/segmentio/analytics-node/pull/113): Flush when the client first receives a message.
* [Internal](link to commit): Remove caching of dependencies on Circle CI.
* [Internal](link to commit): Add License
...
```

f2prateek: (and we can copy this cleaned up version in the github releases section as well)

vadimdemedes: I'm a bit confused here. What is the benefit of having the same release notes in changelog.md and gh releases? Especially when each line in the changelog generated by `git-changelog` links to a corresponding PR.

f2prateek: I don't think the change log file should just be a commit log. It should be more detailed like https://github.com/segmentio/analytics-android/blob/master/CHANGELOG.md.

As for why having it in a file instead of GitHub releases?
1. Less steps to maintain. You edit it during a release, instead of going to the GitHub UI and pasting in the text manually.
2. Easy to search older releases. Grep etc.
3. Version controlled.

Lastly, it gives us the ability to keep a rolling log. The developer making the release doesn't always know details about every change. So as part of each commit/PR, contributors should add a section documenting the public facing version of the change, e.g. migration steps etc. Though this isn't something we're doing diligently at the moment.

Issue #116, opened 2017-07-27
Ability to turn off analytics-node


wlingke:
Because analytics-node is part of our controllers, it would be nice to have the ability to turn it off when we are running our unit tests by passing in an option like:

```
var analytics = new Analytics('YOUR_WRITE_KEY', { off: true });
```

vadimdemedes:
I think this is a valid use case. I'd name this option simply `enable`, rather than `off`.

wlingke:
Awesome! Looking forward to seeing it

wlingke:
@vadimdemedes Hey there - any updates on this??

Thanks!

f2prateek:
@wlingke sorry for the delay, we're about 90% through with the PR but needs some changes still. we'll try get this done soon. Tracking internally at https://segment.atlassian.net/browse/LIB-118

buskerone:
I think this is super important to have when running unit tests. Any updates?

vadimdemedes:
@wlingke @buskerone Would either of you be interested in making a PR for it? https://github.com/segmentio/analytics-node/issues/116#issuecomment-318745025 should be a simple feature.

buskerone:
@vadimdemedes sure if I have enough time I'll do it but for this repo because that's what I'm using right now -> https://github.com/neiker/analytics-react-native 
I'll update u

wlingke:
I'll throw in a PR this weekend. Wish you would've let me known sooner. It
sounded like the segment team was 90% done with this 3 mo ago.

vadimdemedes:
@wlingke You're right, it's our bad, it was 90% done and then I moved from working on this project to internal stuff and abandoned my PR.

wlingke:
Do you still have your pr? If not happy to start from scratch

vadimdemedes:
Yes, here it is: https://github.com/segmentio/analytics-node/pull/118.

wlingke:
Okay thanks. Any particular reason why this didn't get merged in? It looks
good for the most part -- except preventing modification of enabled.

vadimdemedes:
Yep, that's the only reason.

wlingke:
Cool that's easy. I'll get a PR up this weekend.

wlingke:
@vadimdemedes PR made #158

vadimdemedes:
Fixed via #158.

PR #118, opened 2017-07-28
Add `enable` option to disable flushing

vadimdemedes:
Fixes https://github.com/segmentio/analytics-node/issues/116.

Allows to silently skip sending of messages, which is useful in tests and in development. Example usage:

```js
const client = new Client('write key', {
  enable: process.env.NODE_ENV === 'production'
})
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,6 +37,7 @@ class Analytics {
     this.flushAt = Math.max(options.flushAt, 1) || 20
     this.flushInterval = options.flushInterval || 10000
     this.flushed = false
+    this.enabled = typeof options.enable === 'boolean' ? options.enable : true
 
     retries(axios, options.retryCount || 3)
   }
@@ -138,6 +139,10 @@ class Analytics {
   enqueue (type, message, callback) {
     callback = callback || noop
 
+    if (!this.enabled) {
+      return setImmediate(callback)
+    }
+
     message = Object.assign({}, message)
     message.type = type
     message.context = Object.assign({
@@ -187,6 +192,10 @@ class Analytics {
 
   flush (callback) {
     callback = callback || noop
+
+    if (!this.enabled) {
+      return setImmediate(callback)
+    }
 
     if (this.timer) {
       clearTimeout(this.timer)
--- a/test.js
+++ b/test.js
@@ -206,6 +206,18 @@ test('enqueue - extend context', t => {
   t.deepEqual(actualContext, expectedContext)
 })
 
+test('enqueue - skip when client is disabled', async t => {
+  const client = createClient({ enable: false })
+  stub(client, 'flush')
+
+  const callback = spy()
+  client.enqueue('type', {}, callback)
+  await delay(5)
+
+  t.true(callback.calledOnce)
+  t.false(client.flush.called)
+})
+
 test('flush - don\'t fail when queue is empty', async t => {
   const client = createClient()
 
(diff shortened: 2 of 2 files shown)


stephenmathieson:
Duplicate (kinda) of #90 / #75. I'm still unconvinced the shim (see that issue) is worth the added maintenance here.

hulbert:
We're about to add this lib to our systems, and this would be a very nice thing to have. It seems the proposed method of overriding the `flush` method relies on undocumented behavior that could break between versions.

vadimdemedes:
@hulbert is absolutely on point. This is something I'd want too. Recommending overriding `flush()` method is a hackish workaround. I think this option clearly has a practical use case and we need to support one way or another.

Alternatively we could look at `NODE_ENV` and flush only when it's equal to `"production"`. This is what Bugsnag's Node.js library does and I found it a nice convention. Similar to Bugsnag, we would assume environment is production by default, if `NODE_ENV` is empty. If there's a custom way of setting the environment (not setting `NODE_ENV`), we'd have `env` option to override the env name:

```js
new Client('write key', { env: 'production' })
```

Even though it's good, I still prefer having `enable`, because it doesn't make any assumptions about the environment and is explicit. It gives complete control to the user and strips us from maintenance issues with the env way.

vadimdemedes:
@stephenmathieson what do you think?

stephenmathieson:
> we could look at NODE_ENV and flush only when it's equal to "production". 

While this does seem ideal in many cases, it would break our analytics in our non-prod environments. For example, on .build, we set `NODE_ENV=staging`, but we still want to send analytics calls (for testing their implementation, etc).

> we'd have env option to override the env name

I'm 👎 here. Setting `env = "whatever"` seems magical in this case. I'd much prefer something explicit.

> I still prefer having enable, because it doesn't make any assumptions about the environment and is explicit.

Agreed.

> what do you think?

Since there is an obvious demand here, I think this can land. If you fix up the conflicts, I'll approve/merge. 😄

vadimdemedes:
I agree on your argument against env option, @stephenmathieson. Explicit option is better. PR is fixed up and ready to get merged!

index.js
```diff
@@ -54,6 +54,7 @@ class Analytics {
     this.flushAt = Math.max(options.flushAt, 1) || 20
     this.flushInterval = options.flushInterval || 10000
     this.flushed = false
+    this.enabled = options.enable === false ? false : true // eslint-disable-line no-unneeded-ternary
```
stephenmathieson: Strange use of a ternary here (seems the linter agrees with me). Any reason `this.enabled = options.enable` won't work?

vadimdemedes: No idea what happened here 😆

f2prateek:
What if a customer does this:

```
const client = new Client('write key', {
  enable: process.env.NODE_ENV === 'production'
})

client.track(...)
client.enabled = false

client.flush() // either explicitly or as part of a timer 
```

Basically a customer might want to track some events, and then disable the client but still have the expectation of flushing the previously queued events.

I'd prefer if the `enabled` flag was not mutable. I think the JS convention is to prefix it with `_`?

stephenmathieson:
@f2prateek we could just make the `enabled` field immutable so that an error would be raised instead

f2prateek:
SGTM!

stephenmathieson:
@vadimdemedes think you could make the `enabled` field immutable? Something like this should do:

```js
Object.defineProperty(this, 'enabled', {
  configurable: false,
  writable: false,
  value: enabled
})
```

f2prateek:
Closing this for now, feel free to re-open once the PR is updated - otherwise the libraries team will tackle it soon. Tracking here https://segment.atlassian.net/browse/LIB-118

Issue #119, opened 2017-08-17
Method calls hang forever if the underlying request to the Segment server never completes.


serhalp:
If the [request to the `/v1/batch` endpoint](https://github.com/segmentio/analytics-node/blob/4290503572a7c9661661dbcafef75b573cdbda0b/index.js#L225-L237) on the Segment server never completes, the corresponding call to `track()` or `identify()` (etc.) hangs forever and never calls the passed callback with an error / never rejects the promise.

There is currently no way to configure a timeout with this module. In my opinion an abstracted plug-and-play client library for a specific service such as this one should enable retries and timeouts by default and allow configuration.

The underlying request library you're using, superagent, [supports timeouts but they are not enabled by default](http://visionmedia.github.io/superagent/#timeouts). I'll open a PR in a few minutes to allow opting in to timeout functionality. Again, I think it should enabled by default, but that's your call and I'm not sure what the default timeout value should be.

FYI as I understand it the superagent module will retry on timeout errors if it is configured to retry. :+1:

Thank you!

serhalp:
cc @bobzoller @sylspren

stephenmathieson:
If our API isn't responding, we've got huge problems on our hands! Is this something you're running into frequently? If so, please contact our support team!

That said, configurable timeouts/retries sound great to me!

serhalp:
@stephenmathieson Thanks for your concern. It's occurred twice in two weeks. But (back of the envelope) we've likely made on the order of ~200M tracks in that span (from this worker of this service alone). I think 99.9999995% of requests responding eventually is very reasonable. Totally happy with just having the client lib time out and retry 0.000001% of the time. :)

PR #120, opened 2017-08-17
Allow configuring retries and timeout.

serhalp:
- timeout was not enabled nor configurable previously
- retries were hardcoded to `3` and not configurable

Fixes https://github.com/segmentio/analytics-node/issues/119.

I looked into writing tests, but you've got the [mock server set up in a `before`](https://github.com/segmentio/analytics-node/blob/4290503572a7c9661661dbcafef75b573cdbda0b/test.js#L34-L49) so I would need to refactor the tests to allow a test or two to configure the server to conditionally fail with a retryable error or not respond. If the lack of tests is a blocker, I can give it a try.

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,6 +49,8 @@ class Analytics {
     this.queue = []
     this.writeKey = writeKey
     this.host = removeSlash(options.host || 'https://api.segment.io')
+    this.retryCount = options.retryCount || 3
+    this.timeout = options.timeout || false
     this.flushAt = Math.max(options.flushAt, 1) || 20
     this.flushInterval = options.flushInterval || 10000
     this.flushed = false
@@ -225,7 +227,8 @@ class Analytics {
     request
       .post(`${this.host}/v1/batch`)
       .auth(this.writeKey, '')
-      .retry(3)
+      .timeout(this.timeout)
+      .retry(this.retryCount)
       .send(data)
       .end((err, res) => {
         err = err || error(res)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -49,6 +49,8 @@ class Analytics {
     this.queue = []
     this.writeKey = writeKey
     this.host = removeSlash(options.host || 'https://api.segment.io')
+    this.retries = options.retries || 3
```
stephenmathieson: "retries" sounds like boolean to me. Maybe something a bit more descriptive like "retryCount" would be better?

/cc @stevenmiller888

nettofarah: `retryCount` sounds a little less ambiguous to me too.
I'd be down with that.

nettofarah:
Great job, @serhalp.
I still think tests would be pretty critical here before merging.
Do you think you could write at least one test to take care of this scenario.

I was thinking we could add a second condition to the mock server.
something like:

```javascript
if (batch[0] === 'timeout') {
  return setTimeout(() => res.end(), 5000)  
}
```

serhalp:
@nettofarah @stephenmathieson thanks, I renamed `retries` to `retryCount` and added a test for the timeout. Hope you're ok with adding 500 ms to your test suite (5000 ms seemed pretty egregious).

serhalp:
Yeah, but like I was saying with the way the test suite is set up it would be tricky. You need the mock server to respond differently on nth attempt based on an unchanging message. I guess it could use a counter in the closure, but that seems a little messy.

vadimdemedes:
Thank you!

PR #123, opened 2017-10-13
Fix: @angular/cli@1.4.5 Error by adding json file extension

### LINKED ISSUE #122: Integrating into Angular CLI v1.4.5
Seems that the Angular CLI only accepts dropping the file extensions `.ts` and `.js`

The line that causes the build issue is [here](https://github.com/segmentio/analytics-node/blob/4290503572a7c9661661dbcafef75b573cdbda0b/index.js#L9).

A simple fix would be to change this `const version = require('./package').version` to this `const version = require('./package.json').version`.

I know this is outside the scope of this library but thought you all may want to offer Angular CLI support as well :)

### Console Error
```
ERROR in ./node_modules/analytics-node/index.js
Module not found: Error: Can't resolve './package' in '/.../node_modules/analytics-node'
resolve './package' in '/.../node_modules/analytics-node'
  using description file: /.../node_modules/analytics-node/package.json (relative path: .)
    Field 'browser' doesn't contain a valid alias configuration
  after using description file: /.../node_modules/analytics-node/package.json (relative path: .)
    using description file: /.../node_modules/analytics-node/package.json (relative path: ./package)
      no extension
        Field 'browser' doesn't contain a valid alias configuration
        /.../node_modules/analytics-node/package doesn't exist
      .ts
        Field 'browser' doesn't contain a valid alias configuration
        /.../node_modules/analytics-node/package.ts doesn't exist
      .js
        Field 'browser' doesn't contain a valid alias configuration
        /.../node_modules/analytics-node/package.js doesn't exist
      as directory
        /.../node_modules/analytics-node/package doesn't exist
[/.../node_modules/analytics-node/package]
[/.../node_modules/analytics-node/package.ts]
[/.../node_modules/analytics-node/package.js]
[/.../node_modules/analytics-node/package]
 @ ./node_modules/analytics-node/index.js 9:16-36
 @ ./src/app/store/effects/user.effects.ts
 @ ./src/app/store/effects/index.ts
 @ ./src/app/app.module.ts
 @ ./src/main.ts
 @ multi webpack-dev-server/client?http://localhost:4200 ./src/main.ts

webpack: Failed to compile.
```

timdoes:
Fixes #122

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ const validate = require('@segment/loosely-validate-event')
 const request = require('superagent')
 const debug = require('debug')('analytics-node')
 const uid = require('crypto-token')
-const version = require('./package').version
+const version = require('./package.json').version
 
 const setImmediate = global.setImmediate || process.nextTick.bind(process)
 const noop = () => {}


vadimdemedes:
That seems more like an error that belongs to `@angular/cli`, because Node.js natively supports `require()`ing JSON files, without an explicit extension. Did you try searching for existing issues there or opening a new one?

timdoes:
I will create an issue over there tomorrow. I also think it’s more of a TypeScript error due to their tsconfig only allowing ‘.js’ and ‘.ts’ extensions to be dropped.

I agree it should be opened up on their end to us users. Just thought adding the explicit extension would be harmless and a quick fix for all. Thank you for getting back to me as well :)

vadimdemedes:
No worries! I just think it's something that `@angular/cli` is missing and will be a benefit for them to have. You wouldn't want to create PRs for all dependencies that don't use a `.json` extension after all, right? 😆

stephenmathieson:
While I agree, this is 100% an issue on the Angular side, if we can resolve @timdoes's issue by adding a file extension to a require (5 chars), I think it's worth it.

I'm going to look into why the tests are failing and will merge this soon.

timdoes:
Looks like there is a [superagent Large gzip Denial of Service error](https://nodesecurity.io/advisories/479)

stephenmathieson:
@timdoes yep! I saw that too. We've been wanting an excuse to get rid of superagent for a long time now, and this a great one!

See #124 and #126.

I'll merge this right after we get rid of superagent.

stephenmathieson:
Thanks Tim!

Issue #124, opened 2017-10-13
Replace superagent with a different request library


stephenmathieson:
There's been a lot of discussion surrounding this in the past. We need a library that works both client-side and server-side, supports being browserified, doesn't expose globals (`isomorphic-fetch`) and doesn't break the current API.

Currently our tests are failing due to a security vulnerability in superagent:

- https://nodesecurity.io/advisories/479
- https://github.com/visionmedia/superagent/issues/1259

I may be able to own this and get it handled in the next day or so, depending on how urgent we feel this vulnerability is.

Any suggestions for replacement libraries?

/cc @stevenmiller888 @nettofarah @vadimdemedes

vadimdemedes:
Looks like axios is a good fit here! I personally use `got`, but it doesn't have a browser version. Little tip, use `isomorphic-unfetch` instead of `isomorphic-fetch` to save a lot of bytes on the client. It uses extra small https://github.com/developit/unfetch.

nettofarah:
I like axios a lot.
I think their middleware/interceptor feature can help add a lot of cool extra features.

stephenmathieson:
I've seen `isomorphic-unfetch`, it's pretty great!

Sounds like axios is the winner =]

nettofarah:
We can use this sweet library on our tests if we go with axios: https://github.com/nettofarah/axios-vcr

stephenmathieson:
Woah Netto, that's a sweet library! We should hire the guy who wrote it!

nettofarah:
I'll send him this job description: https://segment.com/jobs/525952-Product-Engineer

PR #125, opened 2017-10-13
add `nsprc` to ignore superagent zip bomb vulnerability

stephenmathieson:
Since we're already vulnerable, ignoring this for now while we merge a few PRs and create a new release won't _hurt_ anything. We should obviously address the vulnerability correctly (rather than ignoring it), and finally remove our dependency on superagent (see #124).

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -1,0 +1,3 @@
+{
+  "exceptions": ["https://nodesecurity.io/advisories/479"]
+}


stephenmathieson:
We can close this if #126 lands.

vadimdemedes:
I think https://github.com/segmentio/analytics-node/pull/126 is easy enough to get merged in and better!

PR #126, opened 2017-10-13
replace superagent with axios

### LINKED ISSUE #124: Replace superagent with a different request library

stephenmathieson:
This patch removes our dependency on superagent and replaces it with
axios.

Axios:

- works in the browser
- can be browserified
- works in node
- doesn't leak globals
- has no (known) security vulnerabilities
- is better maintained than superagent

Closes #124.

### DIFF
--- a/index.js
+++ b/index.js
@@ -218,7 +218,7 @@ class Analytics {
     }
 
     const req = {
-      method: 'POSt',
+      method: 'POST',
       url: `${this.host}/v1/batch`,
       data
     }
--- a/index.js
+++ b/index.js
@@ -3,31 +3,15 @@
 const assert = require('assert')
 const removeSlash = require('remove-trailing-slash')
 const validate = require('@segment/loosely-validate-event')
-const request = require('superagent')
+const axios = require('axios')
+const retries = require('axios-retry')
+const ms = require('ms')
 const debug = require('debug')('analytics-node')
 const uid = require('crypto-token')
 const version = require('./package').version
 
 const setImmediate = global.setImmediate || process.nextTick.bind(process)
 const noop = () => {}
-
-/**
- * Get an error from a `res`.
- *
- * @param {Object} res
- * @return {String}
- */
-
-const error = res => {
-  if (!res.error) {
-    return
-  }
-
-  const body = res.body
-  const msg = (body.error && body.error.message) || `${res.status} ${res.text}`
-
-  return new Error(msg)
-}
 
 class Analytics {
   /**
@@ -49,11 +33,13 @@ class Analytics {
     this.queue = []
     this.writeKey = writeKey
     this.host = removeSlash(options.host || 'https://api.segment.io')
-    this.retryCount = options.retryCount || 3
     this.timeout = options.timeout || false
     this.flushAt = Math.max(options.flushAt, 1) || 20
     this.flushInterval = options.flushInterval || 10000
     this.flushed = false
+
+    // setup retries
+    retries(axios, options.retryCount || 3)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -224,20 +210,33 @@ class Analytics {
...
+      method: 'POST',
+      url: `${this.host}/v1/batch`,
+      data
+    }
+
+    if (this.timeout) {
+      req.timeout = typeof this.timeout === 'string' ? ms(this.timeout) : this.timeout
+    }
+
+    axios(req)
```
vadimdemedes: This could be rewritten as:

```js
axios(req)
  .catch(err => {
    if (err.response) {
      err = new Error(err.response.statusText)
    }
    
    return err
  })
  .then(err => {
    callbacks.forEach(callback => callback(err))
    callback(err, data)
  })
```

index.js
```diff
@@ -49,11 +33,13 @@ class Analytics {
     this.queue = []
     this.writeKey = writeKey
     this.host = removeSlash(options.host || 'https://api.segment.io')
-    this.retryCount = options.retryCount || 3
     this.timeout = options.timeout || false
     this.flushAt = Math.max(options.flushAt, 1) || 20
     this.flushInterval = options.flushInterval || 10000
     this.flushed = false
+
+    // setup retries
```
vadimdemedes: I think comment can be removed, since the code is pretty self-explanatory.

stephenmathieson: Good point. Will remove.

index.js
stephenmathieson: Not quite. The furfilled callback receives the "response" from axios (see https://github.com/axios/axios#response-schema). To get this to work, we could do something like:

```
.catch(err => {
  if (err.response) {
    err = new Error(err.response.statusText)
  }
  return err
}).then(err => {
  if (!(err instanceof Error)) {
    err = undefined
  }
  
  callbacks.forEach(callback => callback(err))
  callback(err, data)

  debug('flushed: %o', data)
})
```

But that feels pretty awkward to me.

stephenmathieson: Hmm, could also do:

```js
axios(req)
.then(res => {}) // ignore the response
.catch(err => {
  if (err.response) {
    err = new Error(err.response.statusText)
  }
  return err
})
.then(err => {
  callbacks.forEach(callback => callback(err))
  callback(err, data)

  debug('flushed: %o', data)
})
```

but this also feels a bit strange.

vadimdemedes: You are right, let's leave it like it is in the PR, with a few minor adjustments (will comment separately).

index.js
```diff
@@ -224,20 +209,33 @@ class Analytics {
...
+
+    if (this.timeout) {
+      req.timeout = typeof this.timeout === 'string' ? ms(this.timeout) : this.timeout
+    }
+
+    axios(req)
+    .then(() => done())
+    .catch(err => {
+      if (err.response) {
+        const error = new Error(err.response.statusText)
```
vadimdemedes: => `err = new Error(err.response.statusText)`

stephenmathieson: Maybe it’s just that I’ve been writing a lot of Golang lately (and `govet` yells), but I think shadowing the existing variable decreases readability. I’m open to changing this if you have a strong opinion here tho

vadimdemedes: I hate overwriting variables either, const all the way. Just thought that in this block it wouldn't matter. It can stay that way too, not a big deal.

PR #129, opened 2017-10-18
add useragent

stephenmathieson:
This patch adds the `User-Agent` header to all requests made to the Segment API. In the event we need to track how many requests are coming in from a specific version of the library in the future, we can check CDN logs/stats for this header.

### DIFF
--- a/index.js
+++ b/index.js
@@ -222,7 +222,10 @@ class Analytics {
       auth: {
         username: this.writeKey
       },
-      data
+      data,
+      headers: {
+        'user-agent': `analytics-node@${version}`
+      }
     }
 
     if (this.timeout) {
--- a/index.js
+++ b/index.js
@@ -224,7 +224,7 @@ class Analytics {
       },
       data,
       headers: {
-        'user-agent': `analytics-node@${version}`
+        'user-agent': `analytics-node ${version}`
       }
     }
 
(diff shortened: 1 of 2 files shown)


stephenmathieson:
@nettofarah good point. I'd rather not add this just to tear it out later. Let's discuss further on Slack and hold off merging this.

index.js
```diff
@@ -222,7 +222,10 @@ class Analytics {
       auth: {
         username: this.writeKey
       },
-      data
+      data,
+      headers: {
+        'user-agent': `analytics-node@${version}`
```
vadimdemedes: Perhaps add a space before version for readability?

f2prateek:
I think we should just do UserAgent for now. Private headers don't show up in API logs anyway. I also don't think the format matters for now and it's ok to change later if needed.

I'll add something to https://paper.dropbox.com/doc/analytics-foo-library-guidelines-2trBhLKQD4Soz4VatvuGR eventually.

vadimdemedes:
> I also don't think the format matters for now and it's ok to change later if needed.

I think we can add a space there just for readability sake and making it easier to parse: `const [library, version] = userAgent.split(' ')`. It's a simple enough change to do it now, rather then end up with inconsistent user agents in logs later.

PR #134, opened 2017-11-15
Add an end to end test case.

f2prateek:
Ref: https://paper.dropbox.com/doc/analytics-node-E2E-Test-9oavh3DFcFBXuqCJBe1o9

### DIFF
--- a/test.js
+++ b/test.js
@@ -5,6 +5,9 @@ import delay from 'delay'
 import auth from 'basic-auth'
 import pify from 'pify'
 import test from 'ava'
+import axios from 'axios'
+import retries from 'axios-retry'
+import uid from 'crypto-token'
 import Analytics from '.'
 import {version} from './package'
 
@@ -447,3 +450,50 @@ test('alias - require previousId and userId', t => {
     })
   })
 })
+
+const { RUN_E2E_TESTS } = process.env
+
+if (RUN_E2E_TESTS) {
+  // An end to end to test that sends events to a Segment source, and verifies that a webhook
+  // connected to the source (configured manually via the app) is able to receive the data
+  // sent by this library.
+  // This is described in more detail at https://paper.dropbox.com/doc/analytics-node-E2E-Test-9oavh3DFcFBXuqCJBe1o9.
+  test('end to end test', async t => {
+    const id = uid(16)
+
+    // Segment Write Key for https://segment.com/segment-libraries/sources/analytics_node_e2e_test/overview.
+    // This source is configured to send events to a Runscope bucket used by this test.
+    const analytics = new Analytics('wZqHyttfRO0KxEHyRTujWZQswgTDZx1N')
+    analytics.track({
+      userId: 'prateek',
+      event: 'E2E Test',
+      properties: { id }
+    })
+    analytics.flush()
+
+    // Give some time for events to be delivered from the API to destinations.
+    await delay(5 * 1000) // 5 seconds.
+
+    const axiosClient = axios.create({
+      baseURL: 'https://api.runscope.com',
+      timeout: 10 * 1000,
+      headers: { Authorization: `Bearer ${process.env.RUNSCOPE_TOKEN}` }
+    })
+    retries(axiosClient, { retries: 3 })
+
+    // Runscope Bucket for https://www.runscope.com/stream/zfte7jmy76oz.
+    const messagesResponse = await axiosClient.get('buckets/zfte7jmy76oz/messages?count=10')
+    t.is(messagesResponse.status, 200)
+
+    const requests = messagesResponse.data.data.map(async item => {
+      const response = await axiosClient.get(`buckets/zfte7jmy76oz/messages/${item.uuid}`)
+      t.is(response.status, 200)
+      return JSON.parse(response.data.data.request.body)
+    })
+
+    const messages = await Promise.all(requests)
+
+    const message = messages.find(message => message.properties.id === id)
+    t.truthy(message)
+  })
+}
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,8 @@
 machine:
   node:
     version: 6
+  environment:
+    RUN_E2E_TESTS: true
 
 dependencies:
   pre:


f2prateek:
This is a quick first pass so let me know if you have any suggestions.

I couldn't get `await` working here out of the box (Parsing error: The keyword 'await' is reserved), the code is a bit clunky.

Also if any of the callbacks fail, the test seems to hang forever, likely due to https://github.com/avajs/ava/issues/567.

f2prateek:
Given the number of external systems this calls out to, it'd be nice if there was a way to retry the test if a single run failed, and only fail if say 3/3 runs failed.

Not easily doable with AVA atm.
https://github.com/avajs/ava/issues/934

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
...
+      headers: { 'Authorization': 'Bearer ' + process.env.RUNSCOPE_TOKEN }
+    })
+
+    retries(axiosClient, { retries: 3 })
+
+    axiosClient.get('buckets/zfte7jmy76oz/messages?count=10')
+        .then(function (response) {
+          t.is(response.status, 200)
+
+          // TODO: fail and end the test if none of the requests match.
```
stephenmathieson: We could set a timeout on the client, which would cause an error and result in failure.

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
     })
   })
 })
+
+test.cb('e2e', t => {
+  const id = uid(16)
+
+  const testFn = function () {
```
stephenmathieson: If you make this an `async function`, you’ll be able to `await` promises.

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
...
+    var axiosClient = axios.create({
+      baseURL: 'https://api.runscope.com',
+      timeout: 10 * 1000,
+      headers: { 'Authorization': 'Bearer ' + process.env.RUNSCOPE_TOKEN }
```
stephenmathieson: I’m worried about the implications this will have on open-source contributions, as folks won’t have access to this token.

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
...
+    })
+
+    retries(axiosClient, { retries: 3 })
+
+    axiosClient.get('buckets/zfte7jmy76oz/messages?count=10')
```
stephenmathieson: What is “zfte7jmy76oz”?

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
...
+  // https://segment.com/segment-libraries/sources/analytics_node_e2e_test/overview
+  const analytics = new Analytics('wZqHyttfRO0KxEHyRTujWZQswgTDZx1N', { flushAt: 1 })
+  analytics.track({
+    userId: 'prateek',
+    event: 'E2E Test',
+    properties: {
+      id: `${id}`
+    }
+  }, function () {
+    setTimeout(testFn, 5000)
```
stephenmathieson: What is this timeout for?

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
...
+  })
+  analytics.flush()
```
stephenmathieson: Do we need this `.flush()`? It looks like we tell the library to flush after a single call.

test.js
f2prateek: It's the ID for our Runscope bucket. I'll link to it here.

test.js
f2prateek: Latency between sending to api.segment.io and it being received by an integration.

test.js
f2prateek: Not sure why the timeout on the client would help? The requests won't match if the library had a bug,api was down, etc.

test.js
f2prateek: I think the best we can do here is to run this only on CI? If we don't automate this I'm sure we'll forget to run this test before releasing/making big changes.

test.js
f2prateek: Probably not, I didn't totally trust the library to do the right thing here so I added both.

test.js
f2prateek: And I can't inline the token as it doesn't seem to be very well scoped. Though maybe we could setup a dedicated account for this and it wouldn't matter like we do for SauceLabs.

test.js
f2prateek: I don't think this project is setup for async/await https://github.com/segmentio/analytics-node/pull/134#issuecomment-344448401.

I tried:

```
const response = await axios.get(....)
```

And got the error: "Parsing error: The keyword 'await' is reserved".

stephenmathieson: Right, you need to make the `testFn` async:

```js
const testFn = async function () {
  await somePromise()
}
```

test.js
stephenmathieson: Perhaps moving this test out of the "normal" test suite and _only_ having CI run it would solve the problem?

In `package.json`, we could do:

```json
{
  "scripts": {
    "test": "ava test.js",
    "test-e2e": "ava e2e/test.js
  }
}
```

And in `circle.yml`:

```yml
test:
  override:
    - yarn test
    - yarn test-e2e
```

test.js
stephenmathieson: Ahh ok, the a link would be great. 👍 

Maybe a comment too?

test.js
stephenmathieson: > Currently the test will just hang forever if this happens

I'm suggesting we call `.end()` after a specified amount of time. One way to do this would be to have Axios throw an error after N seconds.

test.js
stephenmathieson: Gotcha. Maybe we should explain that in a comment?

test.js
stephenmathieson: > I didn't totally trust the library

That's not good. Let's chat about that on Slack.

Maybe you should remove it to gain trust in the library? 😉

test.js
vadimdemedes: Instead of creating new folders and files, I propose we leave this test in `test.js` and use [is-ci](https://yarnpkg.com/en/package/is-ci) to determine whether to run it or not.

test.js
vadimdemedes: Could it be shorter?

vadimdemedes:
> Given the number of external systems this calls out to, it'd be nice if there was a way to retry the test if a single run failed, and only fail if say 3/3 runs failed.
> 
> Not easily doable with AVA atm.



You can use the excellent [async-retry](https://yarnpkg.com/en/package/async-retry) for it.

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
     })
   })
 })
+
+test.cb('e2e', t => {
```
vadimdemedes: You should make this test `async` and remove `cb` mode. All of the code in it can be converted to `async/await`.

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
...
+  const id = uid(16)
+
+  const testFn = function () {
+    var axiosClient = axios.create({
```
vadimdemedes: var => const

test.js
vadimdemedes: `headers: { 'Authorization': 'Bearer ' + process.env.RUNSCOPE_TOKEN }` => ```headers: { authorization: `Bearer ${process.env.RUNSCOPE_TOKEN}` }```

test.js
vadimdemedes: better:

```js
const testFn = async () => {

}
```

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
...
+        })
+  }
+
+  // https://segment.com/segment-libraries/sources/analytics_node_e2e_test/overview
+  const analytics = new Analytics('wZqHyttfRO0KxEHyRTujWZQswgTDZx1N', { flushAt: 1 })
+  analytics.track({
+    userId: 'prateek',
+    event: 'E2E Test',
+    properties: {
+      id: `${id}`
```
vadimdemedes: String is pointless here, can just be `{ id }`

test.js
```diff
@@ -447,3 +450,60 @@ test('alias - require previousId and userId', t => {
...
+    }
+  }, function () {
```
vadimdemedes: Use arrow functions.

test.js
stephenmathieson: @vadimdemedes That makes sense for the 90% use case, but what if you _want_ to run this locally?

vadimdemedes: `$ CI=true npm test`

stephenmathieson: Haha good point. Just seems a bit unintuitive to me. That will work for now tho =]

f2prateek: Do we need a dependency for that? Seems like I could just guard it in `process.env.CI === true`?

test.js
f2prateek: Yep I can change that.

test.js
vadimdemedes: Better use `is-ci` to not care about that. All CIs have their specific environment variables, I don't think all of them have `CI` env var.

f2prateek: Seems like an anti pattern to pull in a dependency to replace a small check like this one (a.k.a left-pad). `is-ci` itself is a one liner library, and depends on `ci-info` of which the relevant bits are 7 lines https://github.com/watson/ci-info/blob/master/index.js#L42-L48 (and really only 1-3  lines since we only care about Circle https://github.com/watson/ci-info/blob/master/index.js#L43).

And I think we don't really care about other CI systems? Both Travis/Circle (we're unlikely to use anything else in the near future) have CI=true https://circleci.com/docs/1.0/environment-variables/ https://docs.travis-ci.com/user/environment-variables/.

stephenmathieson: @f2prateek you're thinking in go, not node. In node land, it's 100% acceptable to pull in a 1 line dependency.

vadimdemedes: It's not about lines of code, but not doing something that's already been handled perfectly by other people. Left-pad example doesn't apply anymore, since npm took measures to prevent situations like that in the future. See https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328 for a much more extensive answer to this.

f2prateek: It still doesn't make sense to me to to pull in 50+ lines of dependencies when a one-liner does the job for us. I understand the flexibility is-ci/ci-info will offer, but seems overkill if we don't need that.

f2prateek: But I'll put a pin in in this and use is-ci if that's what you'll want here.

vadimdemedes: As a compromise, I propose we ditch CI check altogether. Instead, let's use something more descriptive of what's actually happening:

```js
if (process.env.INTEGRATION) {
  test(...)
}
```

```
$ INTEGRATION=true npm test
```

There should probably be a better name than `INTEGRATION` too.

stephenmathieson: I very much prefer being explicit here and thought having to `CI=true yarn test` would have been awkward.

test.js
f2prateek: So the async stuff fixed this!

test.js
f2prateek: I left the flush but removed the flushAt:1 part. Can reverse that if we want.

test.js
f2prateek: removed with the async stuff.

test.js
f2prateek: Done.

test.js
f2prateek: Done.

test.js
f2prateek: done.

Should this be a linter check?

test.js
f2prateek: Done.

test.js
```diff
@@ -447,3 +451,50 @@ test('alias - require previousId and userId', t => {
...
+  })
+  retries(axiosClient, { retries: 3 })
+
+  // Runscope Bucket for https://www.runscope.com/stream/zfte7jmy76oz.
+  const messagesResponse = await axiosClient.get('buckets/zfte7jmy76oz/messages?count=10')
+  t.is(messagesResponse.status, 200)
+
+  let found = false
+
+  messagesResponse.data.data.forEach(async item => {
```
stephenmathieson: We could make this a simple `for` loop and only iterate until we "find" what we're looking for:

```js
const {data} = messagesResponse.data
for (const item of data) {
  const response = await axiosClient.get(`buckets/zfte7jmy76oz/messages/${item.uuid}`)
  t.is(response.status, 200)

  const body = JSON.parse(response.data.data.request.body)
  if (id === body.properties.id) {
    found = true
    break
  }
}
```

test.js
```diff
@@ -447,3 +451,50 @@ test('alias - require previousId and userId', t => {
...
+    const response = await axiosClient.get('buckets/zfte7jmy76oz/messages/' + item.uuid)
+    t.is(response.status, 200)
+
+    const body = JSON.parse(response.data.data.request.body)
```
stephenmathieson: I think Axios will return JSON for us if we ask it to.

test.js
```diff
@@ -447,3 +451,50 @@ test('alias - require previousId and userId', t => {
     })
   })
 })
+
+test('e2e', async t => {
```
stephenmathieson: I still think this should only run in certain conditions.

Maybe do something like:

```js
const { RUN_E2E_TESTS } = process.env
if (RUN_E2E_TESTS) {
  test('e2e', async t => { ... })
}
```

And in `circle.yml`, add `RUN_E2E_TESTS=true` in the "environment" config.

stephenmathieson: To avoid nesting, you can do this:

```js
async function e2etest (t) {
  const id = uid(16)
  // ...
}

const { RUN_E2E_TESTS } = process.env
if (RUN_E2E_TESTS) {
  test('e2e', e2etest)
}

Comment (f2prateek) [reply test.js]: Yep, didn't get around to that yet. Will update on the original comment when I do.

File: test.js (hunk 11 above)
Comment (vadimdemedes) [reply test.js]: Yes, it should be that way, because `forEach()` doesn't care about async functions. Meaning, it will not wait until those functions finish.

File: test.js (hunk 13 above)
Comment (vadimdemedes) [reply test.js]: Nesting here is totally fine, looks more natural.

Comment (vadimdemedes) [reply test.js]: @f2prateek could you also modify test title to be more descriptive?

File: test.js (hunk 11 above)
Comment (f2prateek) [reply test.js]: done

File: test.js (hunk 12 above)
Comment (f2prateek) [reply test.js]: How do I get it to do that? I tried:

```
    const response = await axiosClient.get(`buckets/zfte7jmy76oz/messages/${item.uuid}`, {
      responseType: 'json'
    })
```

Comment (vadimdemedes) [reply test.js]: Try passing it in the options when creating `axiosClient`.

File: package.json [hunk 14]
```diff
@@ -51,6 +51,7 @@
     "nsp": "^2.6.3",
     "pify": "^3.0.0",
     "sinon": "^2.3.8",
-    "standard": "^9.0.1"
+    "standard": "^9.0.1",
+    "await-sleep": "^0.0.1"
```
vadimdemedes: Use `delay` instead, it's high quality and widely used.

test.js
```diff
@@ -447,3 +451,51 @@ test('alias - require previousId and userId', t => {
...
+  const id = uid(16)
+
+  // Segment Write Key for https://segment.com/segment-libraries/sources/analytics_node_e2e_test/overview
+  const analytics = new Analytics('wZqHyttfRO0KxEHyRTujWZQswgTDZx1N')
+  analytics.track({
+    userId: 'prateek',
+    event: 'E2E Test',
+    properties: {
```
vadimdemedes: => `properties: { id } `

test.js
```diff
@@ -447,3 +451,51 @@ test('alias - require previousId and userId', t => {
...
+  })
+  analytics.flush()
+
+  // Give some time for events to be delivered from the API to destinations.
+  await sleep(5000)
+
+  const axiosClient = axios.create({
+    baseURL: 'https://api.runscope.com',
+    timeout: 10 * 1000,
+    headers: { 'Authorization': `Bearer ${process.env.RUNSCOPE_TOKEN}` }
```
vadimdemedes: `'Authorization'` => `authorization` (without quotes)

f2prateek: why's that? it should be capitalized?

package.json
f2prateek: Updated.

test.js
```diff
@@ -447,3 +451,51 @@ test('alias - require previousId and userId', t => {
...
+  })
+  retries(axiosClient, { retries: 3 })
+
+  // Runscope Bucket for https://www.runscope.com/stream/zfte7jmy76oz.
+  const messagesResponse = await axiosClient.get('buckets/zfte7jmy76oz/messages?count=10')
+  t.is(messagesResponse.status, 200)
+
+  let found = false
+
+  for (const item of messagesResponse.data.data) {
```
vadimdemedes: It doesn't look good tbh. Much better:

```js
const requests = messagesResponse.data.data.map(async item => {
  const res = await axiosClient.get(`buckets/zfte7jmy76oz/messages/${item.uuid}`)
  return JSON.parse(res.data.data.request.body)
})

const messages = await Promise.all(requests)

const message = messages.find(message => message.properties.id === id)
t.truthy(message)
```

test.js
f2prateek: That should be a thing that should be flagged by the style guide/standard. If it didn't, i don't think it's worth bothering with those changes.

vadimdemedes: It's about consistency. Here it's on multiple lines and on the other ones it's on one (https://github.com/segmentio/analytics-node/pull/134/files#diff-1dd241c4cd3fd1dd89c570cee98b79ddR474, https://github.com/segmentio/analytics-node/pull/134/files#diff-1dd241c4cd3fd1dd89c570cee98b79ddR476).

test.js
f2prateek: Sure, do you have any other suggestions?

test.js
f2prateek: Fixed now.

test.js
vadimdemedes: `send event to runscope`

test.js
f2prateek: Still failed.

```
const axiosClient = axios.create({
      baseURL: 'https://api.runscope.com',
      timeout: 10 * 1000,
      headers: { Authorization: `Bearer ${process.env.RUNSCOPE_TOKEN}` },
      responseType: 'json'
    })
```

```
for (const item of messagesResponse.data.data) {
      const response = await axiosClient.get(`buckets/zfte7jmy76oz/messages/${item.uuid}`)
      t.is(response.status, 200)

      if (id === response.data.data.request.body.properties.id) {
        found = true
        break
      }
    }
```

```
> analytics-node@3.1.1 test /Users/prateek/dev/src/github.com/segmentio/analytics-node
> standard && ava


  30 passed
  1 failed

  end to end test
  /Users/prateek/dev/src/github.com/segmentio/analytics-node/test.js:491

   490:                                                                  
   491:       if (id === response.data.data.request.body.properties.id) {
   492:         found = true                                             

  Rejected promise returned by test. Reason:

  TypeError {
    message: 'Cannot read property \'id\' of undefined',
  }

  Test.<anonymous> (test.js:491:18)
  step (test.js:52:191)
npm ERR! Test failed.  See above for more details.

```

vadimdemedes: Error is different here, `properties` is undefined.

test.js
f2prateek: Updated

test.js
f2prateek: I updated the description and added some comments, let me know if that looks ok.

f2prateek: done

test.js
f2prateek: I think that's okay. The point of code formatter is to not have to make these kinds of changes by hand. If a style guide doesn't flag it, I think it's okay to leave it as creative expression :)

f2prateek: (though I did update it here as I was making other changes at the same time).

test.js
stephenmathieson: HTTP headers are case insensitive.

f2prateek:
I updated some comments to better describe the test for non-Segment contributors.

I would like to add retries but blocked by the errors in https://github.com/segmentio/analytics-node/pull/136. For now I'll merge this as is and can re-prioritize retries in the future if this ends up being a flaky test.

PR #136, opened 2017-11-16
Add retries for end to end test

f2prateek:
Another stab at retries, this seems to be working locally. Sent it as a PR to the branch since that is already reviewed, but I don't want to merge that to master without retries.


f2prateek:
Looks like this fails on node 4 https://circleci.com/gh/segmentio/analytics-node/129?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #137, opened 2017-11-28
Add Size Limit script.

f2prateek:
Currently I'm not enforcing the limit anywhere. This will be just used as a tool to measure our size every once in a while for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,16 @@
   "engines": {
     "node": ">=4"
   },
+  "size-limit": [
+    {
+      "limit": "150 KB",
+      "path": "index.js"
+    }
+  ],
   "scripts": {
+    "size": "size-limit",
     "test": "standard && ava",
+    "testci": "standard && ava && npm run size",
     "prepublish": "npm run check-deps",
     "check-deps": "nsp check"
   },
@@ -52,6 +60,7 @@
     "nsp": "^3.1.0",
     "pify": "^3.0.0",
     "sinon": "^4.1.3",
-    "standard": "^10.0.3"
+    "standard": "^10.0.3",
+    "size-limit": "^0.13.2"
   }
 }
--- a/circle.yml
+++ b/circle.yml
@@ -15,6 +15,6 @@ dependencies:
 test:
   override:
     - npm run check-deps
-    - nvm use 4; npm rebuild &> /dev/null; npm test
-    - nvm use 6; npm rebuild &> /dev/null; npm test
-    - nvm use 8; npm rebuild &> /dev/null; npm test
+    - nvm use 4; npm rebuild &> /dev/null; npm run testci
+    - nvm use 6; npm rebuild &> /dev/null; npm run testci
+    - nvm use 8; npm rebuild &> /dev/null; npm run testci


Rowno:
You could also add it to the CI. 😉

package.json
```diff
@@ -15,7 +15,13 @@
   "engines": {
     "node": ">=4"
   },
+  "size-limit": [
+    {
+      "path": "index.js"
+    }
+  ],
   "scripts": {
+    "size": "size-limit",
```
vadimdemedes: It's useless having this as a separate script, because most likely nobody will run it manually. It should be part of the `test` script. Or better yet, run only on CI.

package.json
```diff
@@ -15,7 +15,13 @@
   "engines": {
     "node": ">=4"
   },
+  "size-limit": [
+    {
+      "path": "index.js"
```
vadimdemedes: I'd set an explicit `limit` here, so that people don't search for `size-limit` docs and look what is the default size limit.

package.json
stephenmathieson: 👍 to running on CI as a separate script

f2prateek: You'll run it separately to get details of which packages are causing the bloat (--why flag).

package.json
f2prateek: I think it's ok not to do this.

vadimdemedes: Why? `size-limit` isn't popular enough so that everyone knows what the defaults are.

f2prateek:
To re-iterate what I mentioned in the original comment, I'm not using this tool to enforce the limit anywhere, but just as a tool to measure the library size (and dependency size) retroactively. 

I see this helpful a development tool and not just as a check. When I add a dependency to try out a new feature, I'd like to be able to quickly measure the impact of adding the dependency, even if the impact is under the arbitrary defined limit we can set.

The one thing enforcing a limit will help with is for external contributions. But afaik this hasn't been a problem in practice so far, so I think it's ok to not add a size limit test to start out.

Fwiw we do this for the android library as well and it's been reasonably helpful.

package.json
f2prateek: Isn't it easier to just say run `npm run size -- --why` than to direct someone to do this from scratch when they need to? I wasted 30 minutes of time trying to figure out how to use this tool. By setting it up here we can prevent this sort of repetitive work.

What's the downside of adding the tool?

package.json
f2prateek: Hmm... my understanding is that there isn't a default size limit and it must be set explicitly (let me know if that's not the case), so no limit is enforced and most folks don't need to know what this tool does or what the defaults are.

As I mentioned in my comments, this isn't being used as a tool to enforce the size limit.

vadimdemedes:
If the limit isn't enforced and it's not affecting the build in any way, I really don't see the point in including it. 

If this is a helpful development tool for you, that's totally fine and you can use it without including it into the repo. Install it globally via `npm i -g size-limit` and you can just run `size-limit` when you want to check the size.

package.json
vadimdemedes: if it's not enforced in the build, this tool is a matter of taste. I, for example, use a totally different tool for checking package sizes, `package-size`. I use it when I need to, but I don't intend to include it in every repo I use it in.

To clear it up about the downside, if it's not taking part in the process of developing this package (part of `test`, ci, etc), it's not part of a repository and just a handy helper for someone.

vadimdemedes: > I wasted 30 minutes of time trying to figure out how to use this tool.

See the latest post below for installing and using it.

f2prateek: Again, the point is when a new developer comes to this repo and is wondering how to do this. It's easier for them to run a npm/make command than figure out the instructions from scratch.

If you want to use package-size, you can still do that, or we could even update this PR to use package-size. This is exactly the point I'm trying to make - a new developer coming into this repo shouldn't have to figure out these things, we should make it easy for them to do this.

vadimdemedes: I don't agree. It's not a complex task to execute, it doesn't require any explanation from us nor interaction with this repo specifically. `size-limit`'s docs are pretty good, their usage section (https://www.npmjs.com/package/size-limit#usage) clearly explains how to use it.

To sum up, we shouldn't include tools that are just for convenience for someone (not everyone) and if they are not affecting the project directly (by taking part in CI build or being part of `npm test`).

f2prateek:
I added size limit to be run on CI.

PR #138, opened 2017-11-28
Stringify userId and anonymousId

f2prateek:
Ref: https://segment.atlassian.net/browse/LIB-29

Our API truncates bignum values. This is problematic because this library has historically accepted Numbers as userIds and anonymousIds. To workaround the truncation, this library needs to stringify userId and anonymousId on the client.

The test cases are numbers that will be truncated in Node, e.g. node -e "console.log(157963456373623802 + 1)".

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ const retries = require('axios-retry')
 const ms = require('ms')
 const uid = require('crypto-token')
 const version = require('./package.json').version
+const isString = require('lodash.isstring')
 
 const setImmediate = global.setImmediate || process.nextTick.bind(process)
 const noop = () => {}
@@ -156,6 +157,16 @@ class Analytics {
 
     if (!message.messageId) {
       message.messageId = `node-${uid(32)}`
+    }
+
+    // Historically this library has accepted strings and numbers as IDs.
+    // However, our spec only allows strings. To avoid breaking compatibility,
+    // we'll coerce these to strings if they aren't already.
+    if (message.anonymousId && !isString(message.anonymousId)) {
+      message.anonymousId = JSON.stringify(message.anonymousId)
+    }
+    if (message.userId && !isString(message.userId)) {
+      message.userId = JSON.stringify(message.userId)
     }
 
     this.queue.push({ message, callback })
--- a/test.js
+++ b/test.js
@@ -141,6 +141,58 @@ test('enqueue - add a message to the queue', t => {
   })
 })
 
+test('enqueue - stringify userId', t => {
+  const client = createClient()
+
+  client.track({
+    userId: 10,
+    event: 'event'
+  }, noop)
+
+  t.is(client.queue.length, 1)
+
+  const item = client.queue.pop()
+
+  t.is(item.message.anonymousId, undefined)
+  t.is(item.message.userId, '10')
+})
+
+test('enqueue - stringify anonymousId', t => {
+  const client = createClient()
+
+  client.screen({
+    anonymousId: 157963456373623802,
+    name: 'screen name'
+  }, noop)
+
+  t.is(client.queue.length, 1)
+
+  const item = client.queue.pop()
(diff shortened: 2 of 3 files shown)


test.js
```diff
@@ -141,6 +141,40 @@ test('enqueue - add a message to the queue', t => {
   })
 })
 
+test('enqueue - stringify userId', t => {
+  const client = createClient()
+
+  client.track({
+    userId: 157963456373623802,
```
vadimdemedes: At the time you pass `157963456373623802` number, it's already rounded down, so stringifying won't preserve the intended value: https://runkit.com/vdemedes/5a1d33234f113b00113d3b4f.

test.js
```diff
@@ -141,6 +141,40 @@ test('enqueue - add a message to the queue', t => {
...
+    event: 'event'
+  }, noop)
+
+  t.is(client.queue.length, 1)
+
+  const item = client.queue.pop()
+
+  t.is(item.message.anonymousId, null)
+  t.is(typeof item.message.userId, 'string')
+  t.regex(item.message.userId, '157963456373623802')
```
vadimdemedes: Should use `t.is` and delete the assertion above, because you're already expecting a string.

index.js
```diff
@@ -8,6 +8,7 @@ const retries = require('axios-retry')
 const ms = require('ms')
 const uid = require('crypto-token')
 const version = require('./package.json').version
+const isString = require('lodash/isstring')
```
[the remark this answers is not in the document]
f2prateek: I don't think that check covers all our bases (e.g. primitive string vs object string). It seems safer to rely on this package, especially as it's already a transitive dependency.

test.js
f2prateek: Ah oops, just copy paste these from another test. Will update.

f2prateek:
@stephenmathieson this only needs to be done for libraries that don't follow the updated spec and all customers to send non-string ids (only ruby, python and node). Others don't allow you to send non string IDs so it's not s problem.

We have to handle this case gracefully for now,  since throwing an error here now will be a breaking change.

index.js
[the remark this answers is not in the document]
vadimdemedes: Checking if a value is a string is standard in all JS projects I've seen. `isString()` from `lodash` and methods alike were useful years ago to help with browser inconsistencies and edge cases.

> primitive string vs object string

Nobody's using "object string", not many people even know it's possible to `new String(x)`. It's best to not optimize for imaginary and weird use cases, after all. Especially when it's clearly wrong code and easily fixable on customer's end, without our involvement. If someone will open up an issue about it (which I highly doubt), we'll deal with it by pointing to a correct usage. We can't optimize for every possible input type. API must have a defined accepted input.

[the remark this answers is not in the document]
f2prateek: Agreed we should be stricter about types. Unfortunately we didn't do this in the past, so for now we must maintain backwards compatibility.

test.js
f2prateek: Updated the test!

test.js
f2prateek: Updated!

f2prateek:
Updated for a final review!

test.js
```diff
@@ -141,6 +141,57 @@ test('enqueue - add a message to the queue', t => {
...
+    anonymousId: 157963456373623802,
+    name: 'screen name'
+  }, noop)
+
+  t.is(client.queue.length, 1)
+
+  const item = client.queue.pop()
+
+  t.is(item.message.userId, undefined)
+  t.is(item.message.anonymousId, '157963456373623800')
```
stephenmathieson: I think a comment explaining why this is expected would be nice. It's not at all intuitive that v8 will round down the number here.

f2prateek: updated

f2prateek:
All good now?

PR #141, opened 2017-12-15
Remove RUN_E2E_TESTS from Circle config.

f2prateek:
Instead of setting this up via the config file, we'll do this as an encrypted variable via the UI https://circleci.com/gh/segmentio/analytics-node/edit#env-vars.

Otherwise the `RUN_E2E_TESTS` variable is true for contributors, but their builds won't have the `RUNSCOPE_TOKEN` which is encrypted and not available to them. This causes contributor PRs to fail as the test suite tries to run the end to end tests without the appropriate tokens.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,8 +1,6 @@
 machine:
   node:
     version: 6
-  environment:
-    RUN_E2E_TESTS: true
 
 dependencies:
   pre:


f2prateek:
(I believe this is what's happening to https://github.com/segmentio/analytics-node/pull/140)

f2prateek:
Also ran into this issue in the Ruby library, and setting both variables as encrypted variables is how we fixed it.

f2prateek:
cc @rohitpaulk

f2prateek:
Digging into CI failures, not sure what's going on there!

f2prateek:
Looks like Circle CI tests are working now!

PR #145, opened 2018-01-10
Only set User-Agent header on browser.

### LINKED ISSUE #139: Refused to set unsafe header "user-agent"
This isn't technically an issue with this library, however it should be noted that neither Chrome nor Safari currently allow setting the "user-agent" header, both throw `Refused to set unsafe header "user-agent"` errors.

It might be responsible to detect whether or not the browser allows for setting this header to avoid spurious errors getting thrown.

I'm happy to open a PR if there's a chance it will go in.  We'd certainly like that, as currently we're getting this error on every page load.

f2prateek:
Closes #139.

### DIFF
--- a/index.js
+++ b/index.js
@@ -220,6 +220,15 @@ class Analytics {
       callback(err, data)
     }
 
+    // Don't set the user agent if we're not on a browser. The latest spec allows
+    // the User-Agent header (see https://fetch.spec.whatwg.org/#terminology-headers
+    // and https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/setRequestHeader),
+    // but browsers such as Chrome and Safari have not caught up.
+    const headers = {}
+    if (typeof window === 'undefined') {
+      headers['user-agent'] = `analytics-node/${version}`
+    }
+
     const req = {
       method: 'POST',
       url: `${this.host}/v1/batch`,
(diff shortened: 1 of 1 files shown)


f2prateek:
Ideally we'd only do this in the browsers that don't support this, but didn't seem like there was a good way to do this yet.

index.js
```diff
@@ -220,16 +220,23 @@ class Analytics {
...
       method: 'POST',
       url: `${this.host}/v1/batch`,
       auth: {
         username: this.writeKey
       },
       data,
-      headers: {
-        'user-agent': `analytics-node/${version}`
-      }
+      headers: headers
```
vadimdemedes: Can just be `headers`, like `data` above.

stephenmathieson: I agree, destructuring this looks a bit nicer

f2prateek: updated.

f2prateek:
thanks for the review @vadimdemedes @stephenmathieson !

index.js
```diff
@@ -220,16 +220,23 @@ class Analytics {
...
     }
 
+    // Don't set the user agent if we're not on a browser. The latest spec allows
+    // the User-Agent header (see https://fetch.spec.whatwg.org/#terminology-headers
+    // and https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/setRequestHeader),
+    // but browsers such as Chrome and Safari have not caught up.
+    const headers = {}
+    if (typeof window === 'undefined') {
+      headers['user-agent'] = `analytics-node/${version}`
+    }
```
sirianni, line 230: The comment and PR title don't match the implementation.  This implementation seems to set the `user-agent` header only if we're **not** on a browser (`window === undefined` means we're not in a browser).

PR #146, opened 2018-01-11
Improve Retry Behaviour.

f2prateek:
Previously we were only retrying network errors or a 5xx error on an idempotent request (GET, HEAD, OPTIONS, PUT or DELETE).

The latter doesn't really guard/apply to us since we use POST for uploading messages. Our POST requests are idempotent since we send messages with a unique message ID that is guarded by our dedupe layer (https://segment.com/blog/exactly-once-delivery/).

For us, our retry policy should be (as per https://paper.dropbox.com/doc/analytics-foo-library-guidelines-2trBhLKQD4Soz4VatvuGR#:uid=189560039280582553736180&h2=Handling-Network-Errors):

1. Retry network errors (socket timed out, etc.)
2. Retry server errors (HTTP 5xx)
3. Retry HTTP 429.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,7 @@ const assert = require('assert')
 const removeSlash = require('remove-trailing-slash')
 const validate = require('@segment/loosely-validate-event')
 const axios = require('axios')
-const retries = require('axios-retry')
+const axiosRetry = require('axios-retry')
 const ms = require('ms')
 const uid = require('crypto-token')
 const version = require('./package.json').version
@@ -38,7 +38,10 @@ class Analytics {
     this.flushInterval = options.flushInterval || 10000
     this.flushed = false
 
-    retries(axios, options.retryCount || 3)
+    axiosRetry(axios, {
+      retries: options.retryCount || 3,
+      retryCondition: this._isErrorRetryable
+    })
   }
 
   /**
@@ -246,6 +249,30 @@ class Analytics {
 
         done(err)
       })
+  }
+
+  _isErrorRetryable (error) {
+    // Retry Network Errors.
+    if (axiosRetry.isNetworkError(error)) {
+      return true
+    }
+
+    if (!error.response) {
+      // Cannot determine if the request can be retried
+      return false
+    }
+
+    // Retry Server Errors (5xx).
+    if (error.response.status >= 500 && error.response.status <= 599) {
+      return true
+    }
+
+    // Retry if rate limited.
+    if (error.response.status === 429) {
+      return true
+    }
+
+    return false
   }
 }
 
--- a/test.js
+++ b/test.js
@@ -503,6 +503,23 @@ test('alias - require previousId and userId', t => {
   })
 })
 
+test('isErrorRetryable', t => {
+  const client = createClient()
+
+  t.false(client._isErrorRetryable({}))
+
+  // ETIMEDOUT is retryable as per `is-retry-allowed` (used by axios-retry in `isNetworkError`).
+  t.true(client._isErrorRetryable({ code: 'ETIMEDOUT' }))
+
+  // ECONNABORTED is not retryable as per `is-retry-allowed` (used by axios-retry in `isNetworkError`).
+  t.false(client._isErrorRetryable({ code: 'ECONNABORTED' }))
+
+  t.true(client._isErrorRetryable({ response: { status: 500 } }))
+  t.true(client._isErrorRetryable({ response: { status: 429 } }))
+
+  t.false(client._isErrorRetryable({ response: { status: 200 } }))
+})
+
 const { RUN_E2E_TESTS } = process.env
 
 if (RUN_E2E_TESTS) {


f2prateek:
hey @vadimdemedes, @stephenmathieson; wanted to get your thoughts on this. I think the behaviour is functionally correct, but not sure on how to test it. https://github.com/softonic/axios-retry doesn't seem to have tests. My initial idea was to test the custom retry function directly (something simple like `t.true(retryCondition({ response: { status: 429 }}))`) but wasn't sure how to access access a  private function in the test.

index.js
```diff
@@ -249,4 +252,26 @@ class Analytics {
   }
 }
 
+var retryCondition = function (error) {
```
vadimdemedes: Should be an arrow function + `const` instead of `var`.

index.js
```diff
@@ -249,4 +252,26 @@ class Analytics {
...
+
+  // Retry Server Errors (5xx).
+  if (error.response.status >= 500 && error.response.status <= 599) {
+    return true
+  }
+
+  // Retry if rate limited.
+  if (error.response.status === 429) {
+    return true
+  }
```
vadimdemedes: This function should return `false` at the end.

index.js
stephenmathieson: I also might suggest adding this method to the Analytics class so we can more easily write tests for it.

f2prateek: what's an arrow function?

f2prateek: Updated to be `const`.

index.js
f2prateek: Updated.

index.js
f2prateek: > I also might suggest adding this method to the Analytics class so we can more easily write tests for it.

Why would that be easier? This seems like a stateless function that can be tested without depending on the analytics class.

Also per my understanding, adding it to the analytics class will automatically expose this function to the public API.

stephenmathieson: In order to test it directly, we'd have to export it. Attaching it directly to `exports` (or `module.exports`) make it _look_ like part of the public API. If we add it as a method of the class, it will be exposed, but won't _look_ like it's part of the public API.

Consider the following:

```js
// exporting directly
exports.somePrivateMethod = () => {}

// exporting as part of the class
module.exports = class Analytics {
  track () {}
  somePrivateMethod () {}
}
```

IMO the first _looks_ like it's exposed for consumption by the end user, while the second just looks like a regular class.

stephenmathieson: > what's an arrow function?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

index.js
```diff
@@ -249,4 +252,28 @@ class Analytics {
   }
 }
 
+const retryCondition = function (error) {
```
stephenmathieson:
```js
const retryCondition = error => {
  // ...
}
```

f2prateek: thanks! roland just explained what this was to me :)

f2prateek:
reminder to myself, don't merge without tests!

index.js
f2prateek: Roland actually suggested not assigning and using it as a function variable at all, which makes sense to me.

```
function _retryCondition(error) {
```

index.js
f2prateek: Thanks! Makes sense about testing, chatted with Roland and he suggested using _ as a prefix to denote it as a private, so I'll do that.

index.js
stephenmathieson: I don't follow, but either way, we need to expose this logic somehow in order to test it.

f2prateek:
Ok, should be ready for review now! Added tests and made some changes from the original approach:

* function moved to Analytics class for testing
* function named with `_` prefix to denote this is not meant to be part of the public API.

@Rowno showed me how I could make the function static too. It wasn't clear if that was fully compatible with what we support today and if it cause issues, so I left it as a plain function for now.

index.js
```diff
@@ -247,6 +250,30 @@ class Analytics {
         done(err)
       })
   }
+
+  _retryCondition (error) {
```
stephenmathieson: Might rename this `shouldRetry` or `isErrorRetryable` for clarity.

test.js
```diff
@@ -503,6 +503,21 @@ test('alias - require previousId and userId', t => {
...
+  t.false(client._retryCondition({}))
+
+  // ETIMEDOUT is retryable as per `is-retry-allowed` (used by axios-retry in `isNetworkError`).
+  t.true(client._retryCondition({ code: 'ETIMEDOUT' }))
+
+  // ECONNABORTED is not retryable as per `is-retry-allowed` (used by axios-retry in `isNetworkError`).
+  t.false(client._retryCondition({ code: 'ECONNABORTED' }))
+
+  t.true(client._retryCondition({ response: { status: 500 } }))
+  t.true(client._retryCondition({ response: { status: 429 } }))
```
stephenmathieson: We should add one last test to verify the default `return false` behavior.

f2prateek: Added one, though it's pretty contrived; we'd never really expect to see and error for  a 2xx response.

index.js
f2prateek: Renamed to `isErrorRetryable`

index.js
vadimdemedes: > Roland actually suggested not assigning and using it as a function variable at all, which makes sense to me.

Why not?

index.js
stephenmathieson: I think you might have forgotten to push?

index.js
Rowno: To me `function _retryCondition(error) {` looks cleaner than `const _retryCondition = error => {`. 🤷‍♂️

index.js
f2prateek: Oops! Pushed now.

index.js
stephenmathieson: 🚲 🏠

PR #148, opened 2018-01-13
Codecov report generation using ava-nyc

suphalb:
This PR will help to generate Codecov Report & Upload it on codecov.io


f2prateek:
hey @suphalb, thanks! can you set this for CI as well? This way the coverage is uploaded to codecov.io (e.g.https://github.com/segmentio/analytics-android/blob/master/.circleci/config.yml#L23)

circle.yml
```diff
@@ -18,3 +18,6 @@ test:
     - nvm use 4; npm rebuild &> /dev/null; npm run testci
     - nvm use 6; npm rebuild &> /dev/null; npm run testci
     - nvm use 8; npm rebuild &> /dev/null; npm run testci
+  post:
+    - npm test
+    - ./node_modules/.bin/nyc report --reporter=lcov > coverage.lcov && bash <(curl -s https://codecov.io/bash)
```
f2prateek: Should this just be `npm run report-coverage` (from https://github.com/segmentio/analytics-node/pull/148/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R30)?

package.json
```diff
@@ -23,10 +23,11 @@
   ],
   "scripts": {
     "size": "size-limit",
-    "test": "standard && ava",
+    "test": "nyc ava tap test.js",
```
f2prateek: We should still be running `standard`.

readme.md
```diff
@@ -28,7 +28,18 @@ client.track({
 
 Documentation is available at [https://segment.com/libraries/node](https://segment.com/libraries/node).
 
+## Generate Codecoverage Report
```
f2prateek: No need to add this information to the Readme.

circle.yml
```diff
@@ -18,3 +18,6 @@ test:
     - nvm use 4; npm rebuild &> /dev/null; npm run testci
     - nvm use 6; npm rebuild &> /dev/null; npm run testci
     - nvm use 8; npm rebuild &> /dev/null; npm run testci
+  post:
+    - npm test
```
f2prateek: Do we need to run `npm test` here? Lines 18-20 are already running a test command as well.

package.json
suphalb: ok i am little confused here . So far i followed https://github.com/segmentio/analytics-node/pull/148#pullrequestreview-89081587  and made it minified / simplified. So will you let me know what should i follow here? I meant should i minify it as per @vadimdemedes request OR it will be same like my previous commit?

circle.yml
suphalb: Yes in fact. But again the same problem i experienced . It works on my local but it seems over circle ci it doesn't generate any report. Ideally any JS inside of package.json should be executed via npm test automatically by circleCi. So to test it out i had to run that cmd individually just to ensure if reports are generating. Now i will make it standard .

package.json
f2prateek: What I mean is we should still be running the `standard` command as part of the test.

I believe @vadimdemedes pointed to https://github.com/sindresorhus/got/blob/master/package.json#L28-L29. From there `xo` = `standard` (both are code formatting tools).

suphalb: ok got it clearly @f2prateek . I am taking care of it quickly and coming back with correct commit.

circle.yml
```diff
@@ -8,7 +8,7 @@ dependencies:
     - nvm install 4
     - nvm install 6
     - nvm install 8
-    - rm -rf node_modules
+    - rm -rf 'node_modules coverage .nyc_output coverage.lcov'
```
stephenmathieson: Since this is wrapped in quotes, the `rm` program will try to delete a file called "node_modules coverage .nyc_output coverage.lcov". I suspect this is not what you had intended.

package.json
```diff
@@ -22,11 +22,8 @@
     }
   ],
   "scripts": {
-    "size": "size-limit",
-    "test": "standard && ava",
-    "testci": "standard && ava && npm run size",
-    "prepublish": "npm run check-deps",
-    "check-deps": "nsp check"
+     "test": "standard && nyc ava tap test.js",
```
stephenmathieson: What is the word "tap" doing here?

readme.md
```diff
@@ -22,13 +22,10 @@ client.track({
   userId: 'user id'
 });
 ```
-
```
stephenmathieson: We ought to keep a newline between these sections of the README.

package.json
suphalb: for TAP format report output as console feedback from cmd?

suphalb: however it's not mandatory but just used for report formation.

stephenmathieson: I think you need to specify it differently if you want AVA to output TAP: https://github.com/avajs/ava#tap-reporter

circle.yml
```diff
@@ -8,7 +8,8 @@ dependencies:
     - nvm install 4
     - nvm install 6
     - nvm install 8
-    - rm -rf 'node_modules coverage .nyc_output coverage.lcov'
+    - rm -rf node_modules ; rm -rf  coverage ; rm -rf .nyc_output ; rm -f coverage.lcov
```
stephenmathieson: No need to separate these with semicolons. The single command will work fine.

suphalb:
So in fact it seems things are looking bit different for me over Circle
CI.  So i tried usual unix style cmd in .yml config such as you mentioned
without semicolons . But strange that it even doesn't execute next
arguement after the first one as it does on traditional nix/unix shell. I
tried it couple of times but on CircleCI console output it doesn't show any
effect. Thus CircleCI simply omitted . So i tried with semicolons as well
as with inv commas.

Second things is same package.json file is being executed without any error
on my local stack but yes as it goes through circleCI it throws some error
and the build fails .

So here i would have to checkout if i am really missing something OR for
Circle CI if i need to use some different parameter/method in config file.



On Wed, Jan 17, 2018 at 3:22 AM, Stephen Mathieson <notifications@github.com

suphalb:
*@vadimdemedes  *so i simplified & minified as you requested as per the
reference , seems that on local it runs without any problem but on CircleCI
the build fails as it seems other additional scripts are necessary for CI.
In fact for codecov line no 2 & last line is suffice but while it's on CI
all other params in scripts section are necessary. I tried 2 builds over CI
through more slimmer version like your given reference. However on my local
it worked flawless but while it's on CI the build failed. So i got back
them again without further refactoring. I would see if there would be any
space to refactor it.

f2prateek:
@suphalb I believe the issue you are seeing is  that locally you are running `npm test` (I guess). On CI, we run `npm run testci` (https://github.com/segmentio/analytics-node/blob/master/circle.yml#L18). In your PR you only updated the command for `npm test`. If you change the `testci` command to be `standard && nyc ava tap test.js && npm run size`; it should work on CI.

package.json
```diff
@@ -22,11 +22,11 @@
     }
   ],
   "scripts": {
-    "size": "size-limit",
-    "test": "standard && ava",
```
f2prateek: @suphalb this was accidentally removed, can you put this back?

package.json
```diff
@@ -22,11 +22,11 @@
...
-    "testci": "standard && ava && npm run size",
-    "prepublish": "npm run check-deps",
-    "check-deps": "nsp check"
+     "size": "size-limit",
```
f2prateek: The indentation is off by 1 here, which makes the diff harder to read. can you leave it as the old indentation?

package.json
suphalb: @f2prateek - can you please consider this latest commit ? https://github.com/segmentio/analytics-node/commit/b0b7874250340d87e37a58f24681076f11cd371e

Seems this time the CI build was successful with the small tweaks you suggested.

vadimdemedes:
@suphalb when you were using an example config I linked to, what was the error on Circle? If it works in other projects, it should as well just work here.

suphalb:
@vadimdemedes - So here it is . Please refer Build test 185 to 188 as following - https://circleci.com/gh/segmentio/analytics-node/185. 

It seems if we consider only codecov and report uploading in fact you are correct that it should work and it was working on my Docker. But it seems for CircleCI there have been additional scripts + test run and those were failing if i remove those additional lines and made it simplify only for codecov. So far all other lines were there from earlier and i didn't refactor it completely. All i just added codecov integration with that previous definition.  So going through @f2prateek's tweak suggestion i corrected and minified a bit and put everything inside of testCI run.

Please let me know if i am missing anything here?

readme.md
```diff
@@ -32,3 +32,4 @@ Documentation is available at [https://segment.com/libraries/node](https://segme
 ## License
 
 Copyright &copy; 2017 Segment Inc. \<friends@segment.com\>
+
```
vadimdemedes: Extra diff.

package.json
```diff
@@ -23,10 +23,10 @@
   ],
   "scripts": {
     "size": "size-limit",
-    "test": "standard && ava",
-    "testci": "standard && ava && npm run size",
+    "testci": "standard &&  nyc ava tap test.js && npm run size",
     "prepublish": "npm run check-deps",
-    "check-deps": "nsp check"
+    "check-deps": "nsp check",
+    "report-coverage": "nyc report --reporter=lcov > coverage.lcov && codecov"
```
vadimdemedes: => `nyc report --reporter=text-lcov | coveralls`

circle.yml
suphalb: @stephenmathieson So removed those semicolons and surprisingly this time CircleCI accepted these traditional unix like cmd. Just bit clueless why it didn't get execute with same thing on my previous build before i added those semicolons and inverted commas .

As your suggestion reverted back the readme with it's original state removing all codecov related text etc.

regarding ava tap i followed the reference from here actually - https://github.com/istanbuljs/nyc.
In fact they referred to use it in same way like i did and i have been directed there by codecov nyc example . So as reference they say - 

```
update the scripts in your package.json to include these bins:
{
   "scripts": {
     "test": "nyc tap ./test/*.js",
     "coverage": "nyc report --reporter=text-lcov > coverage.lcov && codecov"
   }
}
Many testing frameworks (Mocha, Tape, Tap, etc.) can produce TAP output. tap-nyc is a TAP formatter designed to look nice with nyc.
```
I ran without tap param too however it generates the same output but without a nice serial output formation. 

If you think we don't need it here certainly i can omit as it's not mandatory. Please suggest

package.json
```diff
@@ -23,10 +23,10 @@
   ],
   "scripts": {
     "size": "size-limit",
-    "test": "standard && ava",
-    "testci": "standard && ava && npm run size",
+    "testci": "standard &&  nyc ava tap test.js && npm run size",
```
vadimdemedes: Remove `tap` and `test.js`.

circle.yml
```diff
@@ -8,7 +8,8 @@ dependencies:
     - nvm install 4
     - nvm install 6
     - nvm install 8
-    - rm -rf node_modules
+    - rm -rf node_modules coverage .nyc_output coverage.lcov
```
vadimdemedes: No need to remove these new files.

.gitignore
```diff
@@ -1 +1,4 @@
-node_modules
\ No newline at end of file
+node_modules
+coverage
+.nyc_output
+coverage.lcov
```
vadimdemedes: No need to ignore `coverage.lcov`, if you're piping output from `nyc` to `coveralls`, like I mentioned in the comment below.

readme.md
suphalb: Will remove it.

package.json
suphalb: Ok here i am bit confused. I have been informed that we should be using codecov. So am just trying to know will it be codecov integration or coveralls?

package.json
suphalb: Ok as i followed codecov i did it based on their reference here what i replied to @stephenmathieson . https://github.com/segmentio/analytics-node/pull/148#discussion_r162044679

So now i am bit confused about what we are implementing here. Codecov OR Coveralls? 

Just let me get back to @f2prateek here about it?

package.json
vadimdemedes: My mistake, ignore this comment.

package.json
vadimdemedes: My mistake, we're talking about codecov. However, it's still unrelated to this comment.

suphalb: well then i will give it a shot again in same way you suggested. Fact is i tried it for couple of times on my Docker stack but it didn't work so i had to get those back on my local. Once it started to work i replicated same over Cir-CI.

circle.yml
suphalb: mmm am just wondering in anycase if those folder contain some older reports from different commits/codes etc and then we should not remove those before new testbed "PRE" build step so it can fetch/upload only new & updated reports for the new expected commits/changes/codes etc? 

In anycase `report-coverage": "nyc report --reporter=lcov > coverage.lcov && codecov` this will always create and rebuild those dir/files before producing reports & upload them over codecov.

So i am trying to understand here , those dir/files with old reports & meta data should be still there even with the fresh build for new changes?

.gitignore
suphalb: Ok so you mean i should push all of those dir/files with all my local reports & meta data i used for my local test bed? Should not they be ignored like `npm rebuild` as they will be created always while codecov report generation and uploading via nyc process will be running? Please suggest.

package.json
vadimdemedes: It's just `ava tap test.js` is not a valid command, so I'm not sure how it could work for you.

circle.yml
vadimdemedes: Yes, these files will be overwritten in each build anyway, that's why it makes no sense to remove them beforehand.

.gitignore
vadimdemedes: I was talking only about `coverage.lcov`. But since I was thinking about coveralls and piping output before, it can stay there.

package.json
```diff
@@ -23,10 +23,10 @@
   ],
   "scripts": {
     "size": "size-limit",
-    "test": "standard && ava",
```
f2prateek: Don't remove the `test` command (this is still used locally).

suphalb: Rectified.

circle.yml
```diff
@@ -9,6 +9,7 @@ dependencies:
     - nvm install 6
     - nvm install 8
     - rm -rf node_modules
+
```
f2prateek: Was this line added for a reason? Seems unnecessary and inconsistent with the style.

f2prateek:
@vadimdemedes all looks good from me and stephen, wanna give your approval too if this looks ok to you?

package.json
```diff
@@ -23,10 +23,11 @@
...
-    "testci": "standard && ava && npm run size",
+    "test": "standard && nyc ava --tap",
+    "testci": "standard &&  nyc ava --tap && npm run size",
```
vadimdemedes: What's `--tap` for? It doesn't affect coverage or reporting.

vadimdemedes:
LGTM from me after https://github.com/segmentio/analytics-node/pull/148#discussion_r162141910 is resolved.

PR #150, opened 2018-01-16
Replace crypto-token with uuid.

f2prateek:
Both use the node crypto package on Node apps so there shouldn't be a meaningful difference here.

The differences come in the browser. crypto-token works in the browser by polyfilling the crypto package. node-uuid works by using the browser crypto apis if available, otherwise falling back to Math.random() (https://github.com/kelektiv/node-uuid/blob/master/lib/rng-browser.js).

Because of polyfilling on the browser, crypto-token contributes 90kb to our library in the browser which is rougly 80% of the library size on the browser (the rest of the library is only 20kb). node-uuid only adds 1kb (though I believe this comes at the cost of generating slightly less random messageIds on the browser). I don't think the extra weight is worth the cost, especially as we're not using this in a critical security context.

One thing we now do to increase randomness in the browser is includes a hash of the contents of the message as well like we do in a.js (https://github.com/segment-integrations/analytics.js-integration-segmentio/blob/master/lib/index.js#L256). On a side note, analytics.js also uses the uuid package.

Before:

  Package size: 107.79 KB
  With all dependencies, minified and gzipped

After:

  Package size: 13.79 KB
  With all dependencies, minified and gzipped

Because of the significant reduction in size, I also updated the size limit to be 25kb instead of 150kb.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,8 @@ const validate = require('@segment/loosely-validate-event')
 const axios = require('axios')
 const axiosRetry = require('axios-retry')
 const ms = require('ms')
-const uid = require('crypto-token')
+const uuid = require('uuid/v4')
+const md5 = require('md5')
 const version = require('./package.json').version
 const isString = require('lodash.isstring')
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -6,7 +6,8 @@ const validate = require('@segment/loosely-validate-event')
 const axios = require('axios')
 const axiosRetry = require('axios-retry')
 const ms = require('ms')
-const uid = require('crypto-token')
+const uuidv4 = require('uuid/v4')
```
vadimdemedes: Can just be `uuid`. `uuidv4` is ugly.

f2prateek:
@vadimdemedes I just copy pasted that from their docs:

https://github.com/kelektiv/node-uuid

```
const uuidv4 = require('uuid/v4');
uuidv4(); // ⇨ '45db52e1-f95c-4b5f-99a2-8b8d978c99b4'
```

vadimdemedes:
@f2prateek doesn't mean it's the best way to do it :)

f2prateek:
Will change `uuidv4` to `uuid` and rebase after https://github.com/segmentio/analytics-node/pull/148 is merged.

f2prateek:
@vadimdemedes Updated the `uuidv4` usages, and rebased against master. 

Side Note: E2E tests are being flaky because of some destinations downtime today, so if Circle fails I'll kick it off again.

PR #154, opened 2018-01-27
Update loosely-validate-event to 2.0.0

f2prateek:
This version has a breaking change - it'll throw an error for messages > 32kb (as it should, these messages are rejected by the API).

Instead of making a breaking change to analytics-node immediately, we'll have a deprecation period for the old behaviour. During this period, we'll catch this error and log it instead. This will give clients time to update their code as needed.

We may still opt to make a breaking change for analytics-node after this deprecation period, we'll decide this later.


f2prateek:
Closes #152

PR #158, opened 2018-02-24
Add `enable` option to disable flushing

wlingke:
Addresses #116, follow up from #118

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,6 +24,7 @@ class Analytics {
    *   @property {Number} flushAt (default: 20)
    *   @property {Number} flushInterval (default: 10000)
    *   @property {String} host (default: 'https://api.segment.io')
+   *   @property {Boolean} enabled (default: true)
    */
 
   constructor (writeKey, options) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -38,6 +39,12 @@ class Analytics {
     this.flushAt = Math.max(options.flushAt, 1) || 20
     this.flushInterval = options.flushInterval || 10000
     this.flushed = false
+    Object.defineProperty(this, 'enable', {
+      configurable: false,
+      writable: false,
+      enumerable: true,
+      value: options.enable === false ? options.enable : true
```
vadimdemedes: Could you replace that with `typeof options.enable === 'boolean'`?

wlingke:
Isn't what's there and what you're suggesting the same outcome?

vadimdemedes:
Yes, the outcome is the same, but that code reads differently and is quicker to understand its purpose.

wlingke:
Okay updated

vadimdemedes:
@wlingke Thanks!
@f2prateek Could you publish a new minor version? I don't have publish access to this package.

f2prateek:
Released in `3.3.0-beta.2`!

PR #164, opened 2018-04-12
Need to Delay Retry

iwanjunaid:
Sometimes we need to delay retry to avoid high CPU consumes during retry, especially when we set `retryCount` to big number.

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,7 +48,10 @@ class Analytics {
 
     axiosRetry(axios, {
       retries: options.retryCount || 3,
-      retryCondition: this._isErrorRetryable
+      retryCondition: this._isErrorRetryable,
+      retryDelay: function(retryCount) {
+        return options.retryDelay || 500;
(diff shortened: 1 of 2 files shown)


f2prateek:
I think `axiosRetry.exponentialDelay` might be a better option here. https://github.com/softonic/axios-retry

Exponential backoff is generally a good idea and we use that approach in some of the other lirbaries.

f2prateek:
We also don't need to make it an option - let's make exponential retry the default. If folks want control, we can always add it later.

iwanjunaid:
Thank you for you suggestion. I think, yes we should use exponential delay.

iwanjunaid:
Hi @f2prateek What do you think? Is the PR is ok or need some improvements?

iwanjunaid:
Hi guys,

Just wondering when will you guys publish these changes to NPM?

PR #172, opened 2018-05-22
Don't start a timer

### LINKED ISSUE #121: Timer is scheduled to flush even if queue is empty
At the end of the enqueue method a timer is started to flush the queue. However if the queue has already reached the flushAt length this is not necessary.

Even if this behavior is expected we'd like a mechanism to turn off the timer based flushing since we're running on lambda and want to more aggressively flush.

chris-pardy:
Resolves #121 

When the queue has hit the maximum length don't start a timer. If a timer exists flush will cancel it. Timers in node are part of the "event loop" as long as work is in the event loop the node process won't exit, this has impacts to things like AWS Lambda where you're effectively spinning up and shutting down runtimes very quickly.


amirrustam:
This would also be helpful for [bull queue](https://github.com/OptimalBits/bull) workers. Generally it would nice if the batching feature can be disabled via a config option.

uccmen:
we are seeing socket hang up errors when setting `flushAt: 1`. If I understand correctly - this PR will help us resolve this issue as the timer is contributing to the node event loop.

```
[ERROR] Get started analytic { Error: socket hang up
at createHangUpError (_http_client.js:331:15)
at TLSSocket.socketOnEnd (_http_client.js:423:23)
at emitNone (events.js:111:20)
at TLSSocket.emit (events.js:208:7)
at endReadableNT (_stream_readable.js:1064:12)
at _combinedTickCallback (internal/process/next_tick.js:138:11)
at process._tickDomainCallback (internal/process/next_tick.js:218:9)
code: 'ECONNRESET',
```

uccmen:
was this merged?

pbassut:
As @f2prateek pointed out. We needed someone to run this PR on a lambda and check if it doesn't hang.

chris-pardy:
I think I'd closed it since it hadn't gotten merged for several years. Happy to have someone else take it over the finish line though.

test.js
```diff
@@ -254,6 +254,22 @@ test('enqueue - don\'t reset an existing timer', async t => {
...
+  client.enqueue('type', {})
+  t.true(client.flush.calledOnce)
+
+  client.enqueue('type', {})
+  client.enqueue('type', {})
+  t.true(client.flush.calledTwice)
+
+  await delay(10)
+  t.true(client.flush.calledTwice)
+})
```
pbassut, line 271: This test is not clear enough. 
If I understood this correctly: even though it waited 10 seconds it didn't flush a third time because it had already flushed when it reached flushAt.
Maybe the title could be `enqueue - cancel flushing through time interval when already flushed by flushAt`

test.js
```diff
@@ -254,6 +254,22 @@ test('enqueue - don\'t reset an existing timer', async t => {
   t.true(client.flush.calledOnce)
 })
 
+test('enqueue - cancel timer when flushing at max length', async t => {
```
pbassut, line 257:
```suggestion
test('enqueue - prevent flushing through time interval when already flushed by flushAt', async t => {
```

PR #187, opened 2018-08-27
add standardized makefile

f2prateek:
Instead of defining targets here, we delegate to Yarn for most targets.

### DIFF
--- a/.buildscript/bootstrap.sh
+++ b/.buildscript/bootstrap.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+if ! which brew >/dev/null; then
+  echo "homebrew is not available. Install it from http://brew.sh"
+  exit 1
+else
+  echo "homebrew already installed"
+fi
+
+if ! which node >/dev/null; then
+  echo "installing node..."
+  brew install node
+else
+  echo "node already installed"
+fi
+
+if ! which yarn >/dev/null; then
+  echo "installing yarn..."
+  brew install yarn
+else
+  echo "yarn already installed"
+fi
+
+echo "all dependencies installed."
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,8 @@
+bootstrap:
+	.buildscript/bootstrap.sh
+
+# use yarn scripts for the rest
+%:
+	@yarn $@
+
+.PHONY: bootstrap
--- a/package.json
+++ b/package.json
@@ -22,12 +22,14 @@
     }
   ],
   "scripts": {
+    "dependencies": "yarn",
     "size": "size-limit",
     "test": "standard && nyc ava",
     "prepublish": "npm run check-deps",
     "check-deps": "nsp check",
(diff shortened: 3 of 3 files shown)


.buildscript/bootstrap.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+if ! which brew >/dev/null; then
```
stephenmathieson: What if you're not on a Mac? It seems like a pretty arbitrary requirement for working on this package.

.buildscript/bootstrap.sh
```diff
@@ -0,0 +1,24 @@
...
+  echo "homebrew is not available. Install it from http://brew.sh"
+  exit 1
+else
+  echo "homebrew already installed"
+fi
+
+if ! which node >/dev/null; then
```
stephenmathieson: This seems silly/redundant. Users have Node installed already.

.buildscript/bootstrap.sh
```diff
@@ -0,0 +1,24 @@
...
+  echo "installing node..."
+  brew install node
+else
+  echo "node already installed"
+fi
+
+if ! which yarn >/dev/null; then
```
stephenmathieson: Why is Yarn required here? You can "npm run" without Yarn.

Makefile
```diff
@@ -0,0 +1,8 @@
+bootstrap:
+	.buildscript/bootstrap.sh
+
+# use yarn scripts for the rest
+%:
```
stephenmathieson: This just means you now prefix commands with "make" rather than "npm run". I'm wondering why Segment feels this was necessary, since everything seemed to be "OK" before hand.

Makefile
```diff
@@ -0,0 +1,8 @@
...
+	@yarn $@
+
+.PHONY: bootstrap
```
stephenmathieson: PHONY targets usually suggest improper Makefile usage. It's not Grunt, but instead, a tool for compiling things. I think we're misusing it here.

Makefile
f2prateek: The idea here is to have a standard interface for working between libraries. Typically we want `make bootstrap`, `make dependencies`, `make test` to work out of the box (e.g. we have the same targets in the Go library https://github.com/segmentio/analytics-go/blob/v3.0/Makefile).


This is to make it easier for supports agents don't have to hunt down how to run the install step (which can look like `pod install`, `mvn package`, `yarn`, etc.).


As to the implementation, I figured that rather than declare something like this:
```
test:
  yarn run standard && yarn run nyc ava
```

it would be better to re-use the scripts defined in https://github.com/segmentio/analytics-node/blob/master/package.json#L27. And not duplicate ourselves/get out of sync. This also has the advantage that more familiar node users can run `yarn test`, while less experience users with Node may run `make test.

Makefile
f2prateek: 😬 I actually just copied it coz we use it pretty heavily in our Makefiles internally. Happy to remove it here if it doesn't make sense.

PR #190, opened 2018-09-04
Use library-e2e-tester 

f2prateek:
Instead of our handrolled end to end testing code, use library-e2e-tester which is more consistent and supports more message types.

Requires refactoring the existing cli.js file to match the library-e2e-tester interface

### DIFF
--- a/.buildscript/e2e.sh
+++ b/.buildscript/e2e.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+if [ "$RUN_E2E_TESTS" != "true" ]; then
+  echo "Skipping end to end tests."
+else
+  echo "Running end to end tests..."
+  wget https://github.com/segmentio/library-e2e-tester/releases/download/0.2.1/tester_linux_amd64 -O tester
+  chmod +x tester
+  ./tester -path='./cli.js'
+  echo "End to end tests completed!"
+fi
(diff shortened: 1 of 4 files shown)


.buildscript/e2e.sh
```diff
@@ -0,0 +1,13 @@
...
+
+set -e
+
+if [ "$RUN_E2E_TESTS" != "true" ]; then
+  echo "Skipping end to end tests."
+else
+  echo "Running end to end tests..."
+  wget https://github.com/segmentio/library-e2e-tester/releases/download/0.2.1/tester_linux_amd64
+  chmod +x tester_linux_amd64
+  ./tester_linux_amd64 -segment-write-key="$SEGMENT_WRITE_KEY" -webhook-auth-username="$WEBHOOK_AUTH_USERNAME" -webhook-bucket="$WEBHOOK_BUCKET" -path='./cli.js'
```
fathyb: It's an edge case but please note that if `tester_linux_amd64` gets killed by a signal (SIGKILL, SEGSEV, etc) this will leak the value of the arguments. Example with a segfault : 
```
take /tmp/test
echo "void main() { *((int*)(void*)0xdeadbeef) = 1; }" > test.c
gcc test.c -o test -w
./test --write-key=secretstuff
```

You should see `[1]    <PID> segmentation fault  ./test --write-key=secretstuff`.


It'd be safer if the tester could just read the environment variables.

f2prateek: updated to use env vars!

PR #195, opened 2018-10-12
refactor: move cli to a sub-package

dominicbarnes:
The current npm package bundles a CLI, which adds some unnecessary bloat to the package size. This PR removes this bloat, and will be moved to a separate repo/package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,10 +7,7 @@
   "author": {
     "name": "Segment",
     "email": "tools+npm@segment.com",
-    "url": "segment.com"
-  },
-  "bin": {
-    "analytics": "cli.js"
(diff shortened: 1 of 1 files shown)


cli/package.json
```diff
@@ -0,0 +1,45 @@
...
+  ],
+  "keywords": [
+    "analytics",
+    "cli",
+    "segment.io",
+    "segmentio",
+    "segment"
+  ],
+  "dependencies": {
+    "analytics-node": "^3.3.0"
```
f2prateek: Can we make this depend the local version? Otherwise our end to end tests will be testing an outdated version

PR #198, opened 2019-02-25
Removed deprecated nsp

andrenoberto:
Removed deprecated nps and changed to npm audit & fixed dependencies vulnerabilities.

More information about nps can be found below:

NPM blog post: https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting
NPM Docs: https://docs.npmjs.com/auditing-package-dependencies-for-security-vulnerabilities

I've changed 
`nsp check`
to 
`npm audit --audit-level high`

Also I've created a `npm-shrinkwrap.json` file, since it's required by npm audit command (another option is to use package-lock.json).

To create this file I've just run `npm shrinkwrap`

Note that the following files have been updated:
```
package.json
yarn.lock
```
The reason for that is I've changed the versions of some dependencies (the ones that had any level of vulnerability found by npm audit).

I also removed nsp dev-dependency as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,14 @@ jobs:
     <<: *node-base-test
     docker:
       - image: circleci/node:6-browsers
+    steps:
+      - run:
+          name: "Update npm"
+          command: |
+            npm install npm@latest
+            sudo rm -rf /usr/local/lib/node_modules/npm
+            sudo mv node_modules/npm /usr/local/lib/node_modules/npm
+            sudo chown -R 500:500 /usr/local/lib/node_modules/npm
   test-node8:
     <<: *node-base-test
     docker:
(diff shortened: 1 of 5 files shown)


andrenoberto:
FYI: I'll check why build is failing as soon as I get home.

andrenoberto:
Alright, after updating the tests I was able to pass most of CircleCI environments but there is a trade off for using npm audit. It needs npm version to be > 6.4.0. Do you guys have any idea on how to fix this?

andrenoberto:
Alright guys, so I got the node6 environment fixed by manually updating only the npm version. 
But there is still a job that I can't fix, which is snyk case.

`snyk test requires an authenticated account. Please run `snyk auth` and try again.`

So, I guess you might have to do something on your side, not quite sure as I'm not that experienced snyk guy.

I'll be waiting for your feedback.

f2prateek:
I just removed it - closed by #200

PR #200, opened 2019-03-25
fix builds

f2prateek:
* remove nsp check as it is deprecated (https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting). we already have a replacement with Snyk.
* upgrade handlebar dependency (https://snyk.io/vuln/SNYK-JS-HANDLEBARS-173692)


f2prateek:
closes #198

PR #216, opened 2019-06-20
fix context declaration

f2prateek:
context is declared on the workflow, rather than the job.

```
> circleci config validate -c .circleci/config.yml

.circleci/config.yml is valid
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ workflows:
       - test-node10
       - test-node12
   test_and_publish:
+    context: npm
     jobs:
       - test:
           filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -6,6 +6,7 @@ workflows:
       - test-node10
       - test-node12
   test_and_publish:
+    context: npm
```
nickatsegment: The context goes on the job instance in the workflow:
```
workflows:
  version: 2
  my-workflow:
    jobs:
      - run-tests:
          context: org-global
```

f2prateek: 🤦‍♂ updating

PR #225, opened 2019-08-23
add pre-commit circleci test

danieljackins:
https://segment.atlassian.net/browse/LIB-1137

Adds pre-commit validation of .circleci/config.yml file. Requires circleci local CLI to be installed https://circleci.com/docs/2.0/local-cli/

### DIFF
--- a/.huskyrc.js
+++ b/.huskyrc.js
@@ -1,0 +1,14 @@
+const tasks = arr => arr.join(' && ')
+
+module.exports = {
+  'hooks': {
+    'pre-commit': tasks([
+      'exec < /dev/tty',
+      `if ! eMSG=$(circleci config validate -c .circleci/config.yml); then
+        echo "CircleCI Configuration Failed Validation." 
+        echo $eMSG 
+        exit 1
+       fi`
+    ])
+  }
+}
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "@segment/loosely-validate-event": "^2.0.0",
     "axios": "^0.19.0",
     "axios-retry": "^3.0.2",
+    "husky": "^3.0.4",
     "lodash.isstring": "^4.0.1",
     "md5": "^2.2.1",
     "ms": "^2.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1673,6 +1673,11 @@ ci-info@^1.5.0:
   resolved "https://registry.yarnpkg.com/ci-info/-/ci-info-1.6.0.tgz#2ca20dbb9ceb32d4524a683303313f0304b1e497"
   integrity sha512-vsGdkwSCDpWmP80ncATX7iea5DWQemg1UgCW5J8tqjU3lYw4FBYuj89J0CTVomA7BEfvSZd84GmHko+MxFQU2A==
 
+ci-info@^2.0.0:
+  version "2.0.0"
+  resolved "https://registry.yarnpkg.com/ci-info/-/ci-info-2.0.0.tgz#67a9e964be31a51e15e5010d58e6f12834002f46"
+  integrity sha512-5tK7EtrZ0N+OLFMthtqOj4fI2Jeb88C4CAZPu25LDVUgXJ0A3Js4PMGqrn0JU1W0Mh1/Z8wZzYPxqUrXeBboCQ==
+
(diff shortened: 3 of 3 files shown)


danieljackins:
Should this automatically install circleci local CLI as well or should that be left up to the users?

.huskyrc.js
```diff
@@ -0,0 +1,14 @@
+const tasks = arr => arr.join(' && ')
+
+module.exports = {
+  'hooks': {
+    'pre-commit': tasks([
+      'exec < /dev/tty',
+      `if ! eMSG=$(circleci config validate -c .circleci/config.yml); then
```
f2prateek: Can we print a nicer error message if the circleci tool doesn't exist?

f2prateek:
Can we also run the validation step on CI itself, and fail loudly with an error? Previously, the CI job would continue even if the config is invalid.

https://circleci.com/gh/segmentio/analytics-node/21384

jobs: publish: extraneous key [context] is not permitted

.huskyrc.js
danieljackins: bash printed out a message for it, is it not enough?
`Daniels-MacBook-Pro:analytics-node daniel.jackins$ git commit -m "test"
husky > pre-commit (node v8.16.0)
/bin/sh: circleci: command not found
CircleCI Configuration Failed Validation.

husky > pre-commit hook failed (add --no-verify to bypass)`

f2prateek: Since we are making having the tool installed locally a requirement, I think ideally the message should print something like "Could not find `circleci` tool, please install it from https://circleci.com/docs/2.0/local-cli/".

As a new person contributing to this repo (especially external contributors), I would have no idea what this tool is. I would also okay with not validating this locally and forcing installing the CLI tool, as long as the CI is validating it. This would make it friendlier for new contributors, as they would not not have to go out of their way to install the CLI tool.

danieljackins: It could be kind of nice to have the local validation as well to catch the errors before you've pushed the change and wait for circleci to catch it, though. Already saved me from pushing a change that had a tab instead of spaces :)

But if the hassle outweighs the benefit I can take out the local validation

danieljackins: (and they could always use the --no-verify if they just want to skip the local validation)

package.json
```diff
@@ -42,6 +42,7 @@
     "@segment/loosely-validate-event": "^2.0.0",
     "axios": "^0.19.0",
     "axios-retry": "^3.0.2",
+    "husky": "^3.0.4",
```
f2prateek: This should be only a development dependency.

.huskyrc.js
```diff
@@ -0,0 +1,16 @@
+const tasks = arr => arr.join(' && ')
+module.exports = {
+  'hooks': {
+    'pre-commit': tasks([
+      'exec < /dev/tty',
+      `if ! [ -x "$(command -v circleci2)" ]; then
```
f2prateek: why is this looking for the `circleci2` command, but then using the `circleci` command?

.circleci/config.yml
```diff
@@ -49,6 +49,17 @@ workflows:
 
 version: 2
 jobs:
+  validate-config:
+    docker:
+      - image: node
+    steps:
+      - run:
+        - >
+          if ! eMSG=$(circleci config validate -c .circleci/config.yml); then
```
f2prateek: We should share the command between this and the husky commit hook so we're not repeating ourselves.

f2prateek:
I think we should add a new NPM script (here https://github.com/segmentio/analytics-node/blob/master/package.json#L21) `circle-lint`. The `circle-lint` command will validate that the circle CI config is validate, and fail if the validation does not succeed.

There will be three ways to run this:
1. a developer can run manually `yarn run circle-lint`.
2. the CI should also run `yarn run circle-lint`.
3. the git hook can run `yarn run circle-lint`.  Since this is automatically run on a git hook and requires an extra install step, only for this one we should verify if the CLI tool is available or not, and show a message with a link to to install if a developer would like to lint this as part of a pre-commit hook.

.huskyrc.js
danieljackins: oops, was making sure it worked properly for invalid commands. thanks for catching that

Issue #230, opened 2019-09-23
Data parameter does not return any data when calling .track method


albertmourato:
When calling `.track` method, it may receive a callback function that has as parameters `(error, data)`. `error` is coming `undefined` which is good, since my code was executed successfully, however, `data` always comes `undefined` which is not good at all.

yuriedmundo:
I have the same problem

Chathula:
same issue here. I checked the code. actually it doesn't return anything as I see :(

pbassut:
We're looking into it. PR looks good. Will have updates soon

PR #231, opened 2019-09-23
Fix data parameter always undefined on callback for track method

albertmourato:
Fixes: https://github.com/segmentio/analytics-node/issues/230

### DIFF
--- a/index.js
+++ b/index.js
@@ -210,7 +210,7 @@ class Analytics {
     }
 
     if (this.queue.length >= this.flushAt) {
-      this.flush()
+      this.flush(callback)
     }
 
     if (this.flushInterval && !this.timer) {
--- a/index.js
+++ b/index.js
@@ -214,7 +214,7 @@ class Analytics {
     }
 
     if (this.flushInterval && !this.timer) {
-      this.timer = setTimeout(this.flush.bind(this), this.flushInterval)
+      this.timer = setTimeout(this.flush.bind(this, callback), this.flushInterval)
     }
   }
 
--- a/index.js
+++ b/index.js
@@ -205,7 +205,7 @@ class Analytics {
 
     if (!this.flushed) {
       this.flushed = true
-      this.flush()
+      this.flush(callback)
       return
     }
 


albertmourato:
@danieljackins I didn't need to change anywhere else since the place I added solved my problem. Feel free to add in the other places.

albertmourato:
@danieljackins just added the callback parameter to the `this.flush` call as you asked. Could you please review this PR again?
There are other people facing the same problem.
Check https://github.com/segmentio/analytics-node/issues/230

index.js
```diff
@@ -205,12 +205,12 @@ class Analytics {
...
+      this.flush(callback)
       return
     }
 
     if (this.queue.length >= this.flushAt) {
-      this.flush()
+      this.flush(callback)
     }
 
     if (this.flushInterval && !this.timer) {
```
pbassut, line 216: There's also another one here:
https://github.com/segmentio/analytics-node/blob/d2c565d27110b8c3846372a8d8aacc69b0ef83ca/index.js#L217

pbassut (changes requested):
Mind make that little change? Just so we're consistent on all three callback

index.js
albertmourato: thanks, @pbassut
just applied the suggestion!

nd4p90x (approved):

index.js
```diff
@@ -213,16 +213,16 @@ class Analytics {
...
     if (!this.flushed) {
       this.flushed = true
-      this.flush()
+      this.flush(callback)
       return
     }
 
     if (this.queue.length >= this.flushAt) {
-      this.flush()
+      this.flush(callback)
```
Dahaden, line 221: Weird bug, may not cause issues, but this actually causes the callback to be called twice now as its the exact same callback pulled from the item when flushed.
https://github.com/segmentio/analytics-node/blob/master/index.js#L247-L259

pooyaj: cc - @pbassut

BrandonNoad: Yes, the callback is running twice for me as well. Is there a plan to fix that? Or should I log a new issue?

pbassut: We should have a fix ASAP

BrandonNoad: Hi. Was the fix for this released with v5.1.0?

pooyaj: @pbassut could you please check if this is already fixed?

pbassut: @pooyaj It's fixed locally. I'm writing some tests and I'll create a PR for it.

pbassut: @BrandonNoad Can you give me a minimal working example that illustrates the issue? 

For issue `#1`: seems unlikely given they're only called when the request to segment servers actually finishes(successfully or not).

As for issue `#2`: I'll have to see some code.

pbassut: @BrandonNoad Please open a new issue for it though.

BrandonNoad: @pbassut Yes, I can open a new issue with some code to reproduce the behaviour.

> For issue #1: seems unlikely given they're only called when the request to segment servers actually finishes(successfully or not).

I may have been incorrect in saying "too early". That was just an assumption based on the fact that the callback is being called twice and in the first call, `batch` is `undefined`.

BrandonNoad: New issue is here: https://github.com/segmentio/analytics-node/issues/300

PR #240, opened 2020-03-12
Feature: Allow passing axios instance or config in options.

krisdages:
### Background:

I have an Electron app using axios for HTTP where I am using a custom adapter to get around proxy issues by using Electron's `net` module.

### My workaround:
The current version of `analytics-node` doesn't expose any configuration for `axios`, so I added the following `options` properties:
- `axiosConfig`: configuration options for creating an axios instance if you don't have axios as a dependency
- `axiosInstance`: instance of axios if you already have axios as a dependency, maybe with a different version

It's not ideal from a library-coupling perspective, but I didn't see an easy way to avoid that without 
defining some library-agnostic interface for the HTTP adapter which would still support retries, which seems overkill.

I just thought I would post this PR in case it was a useful enhancement. 

### Alternative considered:
One option `axios` with either a config object or a function returning an axios instance.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,10 +21,12 @@ class Analytics {
    *
    * @param {String} writeKey
    * @param {Object} [options] (optional)
-   *   @property {Number} flushAt (default: 20)
-   *   @property {Number} flushInterval (default: 10000)
-   *   @property {String} host (default: 'https://api.segment.io')
-   *   @property {Boolean} enable (default: true)
+   *   @property {Number} [flushAt] (default: 20)
+   *   @property {Number} [flushInterval] (default: 10000)
+   *   @property {String} [host] (default: 'https://api.segment.io')
+   *   @property {Boolean} [enable] (default: true)
+   *   @property {Object} [axiosConfig] (optional)
+   *   @property {Object} [axiosInstance] (default: axios.create(options.axiosConfig))
    */
 
   constructor (writeKey, options) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -21,10 +21,12 @@ class Analytics {
...
    * @param {String} writeKey
    * @param {Object} [options] (optional)
-   *   @property {Number} flushAt (default: 20)
-   *   @property {Number} flushInterval (default: 10000)
-   *   @property {String} host (default: 'https://api.segment.io')
-   *   @property {Boolean} enable (default: true)
+   *   @property {Number} [flushAt] (default: 20)
+   *   @property {Number} [flushInterval] (default: 10000)
+   *   @property {String} [host] (default: 'https://api.segment.io')
+   *   @property {Boolean} [enable] (default: true)
```
pbassut, line 27: Thanks for those updates

pbassut (approved):
If we merge this we can work out https://github.com/segmentio/analytics-node/pull/246 by using axios' baseURL parameter(and also leaving feature like proxy intact).

We can merge this and I'll tackle that later. Leaving this here for reference

pooyaj (approved):

pooyaj:
@krisdages @pbassut could you please fix the conflict so we can merge this?

pbassut:
Sure. Notice we don't need to merge https://github.com/segmentio/analytics-node/pull/246 if we merge this.
I would even prefer to merge this as this provides more flexibility.

pbassut:
Also, I'll have to create another PR to fix the conflicts here.

pooyaj:
@pbassut oh, sorry just merged the other one. Agree, this would've covered that.

pbassut:
https://github.com/segmentio/analytics-node/pull/270

PR #246, opened 2020-07-16
Allow to update endpoint to send batch events

ankurchandraa:
When initializing the SDK, allows the developer to have a customized path for sending batch events

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,6 +35,7 @@ class Analytics {
     this.queue = []
     this.writeKey = writeKey
     this.host = removeSlash(options.host || 'https://api.segment.io')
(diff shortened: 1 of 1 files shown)


pbassut:
Thanks for this!
we're tracking this internally and we'll merge it soon

pbassut (approved):

nd4p90x (approved):

pooyaj:
@ankurchandraa I'm curios what is the use-case for changing the name of the batch endpoint?

pbassut:
In case you're using a proxy then you're don't need to sticky to that endpoint name.

Issue #239, opened 2020-03-10
Infinite loop due to axios 0.19.x and axios-retry


Floriferous:
Issues have been opened in `axios` and `axios-retry`, as a combination of `axios` 0.19.x and `axios-retry` 3.1.x are causing this package to end up in an infinite loop.

Relevant issues:
* https://github.com/axios/axios/issues/2203
* https://github.com/softonic/axios-retry/issues/110

Some people are reporting that the issue is gone for them in `0.19.2`, but not on our end.

pbassut:
Fixed at https://github.com/segmentio/analytics-node/pull/255

Issue #223, opened 2019-08-14
axios-retry package is problematic


broom9:
It's currently in the dependency but it hangs on network errors. Please consider to use an alternative solution.

dhakalsuresh:
it's takes me forever to figure out that this package hangs on network errors and consumes all CPU

g-guirado:
axios-retry does not work with axios 0.19.0 (https://github.com/softonic/axios-retry/issues/59, also mentioned on https://github.com/softonic/axios-retry/blob/master/README.md).

Why does an analytics package use axios-retry? IMHO analytics should be best effort and dropped on error, no need to retry (or at least retry shouldn't be the default behavior)

Floriferous:
Analytics should definitely not be dropped! We rely on our analytics as a high-trust datasource, for some of our low-volume events, we don't want a single one of them to be dropped.

pooyaj:
Closing since there has been few updates to axios since then. Please feel free to re-open if still an issue

PR #255, opened 2020-10-19
Fix for infinite axios-retry on 5xx responses and axios client option leakage

yujidude:
This merge seeks to address two issues:

1. Fix the infinite retry bug associated with axios 0.19.0
2. Prevent axios-retry options from leaking into other codebases that are using axios

To address the first concern I've updated the current version of axios to 0.19.2 which contains the necessary fixes for avoiding infinite retries within axios when the HTTP response is within the 5xx range. 

As per the second concern, I've configured the Analytics class to instead use it's own axios client, thereby preventing client options from leaking into other axios instances. 

Unit tests have been added to validate that the retries are in fact working and that option leakage is no longer occuring.

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,8 +45,8 @@ class Analytics {
       enumerable: true,
       value: typeof options.enable === 'boolean' ? options.enable : true
     })
-
-    axiosRetry(axios, {
+    this.axiosClient = axios.create()
+    axiosRetry(this.axiosClient, {
       retries: options.retryCount || 3,
       retryCondition: this._isErrorRetryable,
       retryDelay: axiosRetry.exponentialDelay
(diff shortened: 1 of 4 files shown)


yujidude:
This pull request resolves #239 and #223.

yujidude:
Hi @ivolo @travisjeffery @f2prateek  - I noticed that the three of you appear to be the top committers to this repository and I was interested in getting your thoughts on this particular pull request. I believe this PR will resolve a few outstanding issues and I'd love to help with a potential fix. Is there anything additional that you'd like to see in this PR for it to receive approval?

f2prateek:
cc @lubird @pooyaj

pooyaj (approved):
Thanks for sending the PR! Looks great ✨

pooyaj:
@yujidude I published `3.4.0-beta.3` to npm!

yujidude:
Fantastic! Thank you again for the review!

PR #273, opened 2021-02-26
upgrade uuid dep to support bundler tree shaking

gwu:
See https://github.com/uuidjs/uuid/blob/master/README_js.md#deep-requires-now-deprecated

From the above document, the motivation for upgrading:

As of uuid@7.x this library now provides ECMAScript modules builds,
which allow packagers like Webpack and Rollup to do "tree-shaking" to
remove dead code.


gwu:
I'm not sure how to get the "ci/circleci: synk" check to run. Do I need an account or could somebody on the segmentio side help make sure it passes?

pooyaj (approved):

Issue #287, opened 2021-08-02
Promise returned by flush never rejects (and possible tests that do not work properly)


Dahaden:
Just noticed the latest changes includes the promise being returned by flush. Love it!

However, after looking at the code, I noticed that the catch swallows the error and then the promise is resolved with `undefined`.

This is easily testable by running something like

```
Promise.reject('test').catch(e =>
    console.log(e)
).then(() =>
    console.log('I end up in console.')
).catch(e =>
    console.log('I will not be logged.')
)
```

A fix to this is either:

1. `return Promise.reject(error)` in the catch block, or
2. `throw error` also inside the catch block

One big caveat to doing this though is that it may cause consumers to see `Unhandled rejection` errors if they do not catch these promises themselves.

I then had a look at the tests to see how to test this and saw https://github.com/segmentio/analytics-node/blob/master/test.js#L334-L346

```

test('flush - respond with an error', async t => {
  const client = createClient()
  const callback = spy()

  client.queue = [
    {
      message: 'error',
      callback
    }
  ]

  await t.throws(client.flush(), 'Bad Request')
})
```
which appears to test this exact functionality (even before the flush function returned a promise!)

I noticed that the dependency `ava` is quite out of date and bumped it to 3.15.0 and almost got everything working, but am now getting `Unhanlded rejection` errors.

Will push my branch up in a PR to show what I have done, but no idea how to fix this.

muhammadmahindar:
@Dahaden can you let me know if this will be fixed soon? I am facing the same problem.

Dahaden:
@muhammadmahindar Sorry almost forgot about this. Found what was causing me all the issues. New PR up with less changes in it
https://github.com/segmentio/analytics-node/pull/294

muhammadmahindar:
@Dahaden any timeline for when this PR will be released?

Dahaden:
I dont work for Segment or have any permissions on this repo.

I have made a comment suggesting that this could be considered a major version bump due to the change in `client.flush` actually rejecting promises and the issue of causing some consumers to get many warnings or even [throws errors](https://developer.ibm.com/blogs/nodejs-15-release-blog/).

muhammadmahindar:
@Dahaden Thanks a lot dude. I think I'll have to clone until this is fixed.

PR #288, opened 2021-08-02
#287 Dhaden/flush return rejectable promise

Dahaden:
Referencing https://github.com/segmentio/analytics-node/issues/287

- Return rejected promises via flush
- Attempted to fix tests (not quite there)

### DIFF
--- a/test.js
+++ b/test.js
@@ -1,12 +1,12 @@
-import { spy, stub } from 'sinon'
-import bodyParser from 'body-parser'
-import express from 'express'
-import delay from 'delay'
-import auth from 'basic-auth'
-import pify from 'pify'
-import test from 'ava'
-import Analytics from '.'
-import { version } from './package'
+const { spy, stub } = require('sinon')
+const bodyParser = require('body-parser')
+const express = require('express')
+const delay = require('delay')
+const auth = require('basic-auth')
+const pify = require('pify')
+const test = require('ava')
+const Analytics = require('./index.js')
+const { version } = require('./package')
 
 const noop = () => {}
 
@@ -88,7 +88,8 @@ test('expose a constructor', t => {
 })
 
 test('require a write key', t => {
-  t.throws(() => new Analytics(), 'You must pass your Segment project\'s write key.')
+  const error = t.throws(() => new Analytics(), { code: 'ERR_ASSERTION' })
+  t.is(error.message, 'You must pass your Segment project\'s write key.')
 })
 
 test('create a queue', t => {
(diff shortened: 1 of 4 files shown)


test.js
```diff
@@ -1,12 +1,12 @@
-import { spy, stub } from 'sinon'
-import bodyParser from 'body-parser'
-import express from 'express'
-import delay from 'delay'
-import auth from 'basic-auth'
-import pify from 'pify'
-import test from 'ava'
-import Analytics from '.'
-import { version } from './package'
+const { spy, stub } = require('sinon')
```
Dahaden, line 1: Annoyingly bable was removed from ava and imports arent native until node 13 so have to revert to requires here

test.js
```diff
@@ -88,7 +88,8 @@ test('expose a constructor', t => {
 })
 
 test('require a write key', t => {
-  t.throws(() => new Analytics(), 'You must pass your Segment project\'s write key.')
+  const error = t.throws(() => new Analytics(), { code: 'ERR_ASSERTION' })
+  t.is(error.message, 'You must pass your Segment project\'s write key.')
```
Dahaden, line 92: With latest versions of ava, you can no longer provide just a string.

I can make this consistent with the other tests, or use this method for testing exceptions. Either or.

test.js
```diff
@@ -296,7 +297,7 @@ test('enqueue - skip when client is disabled', async t => {
 test('flush - don\'t fail when queue is empty', async t => {
   const client = createClient()
 
-  await t.notThrows(client.flush())
+  await t.notThrowsAsync(() => client.flush())
```
Dahaden, line 300: For anything that returns a promise, we have to use `t.throwsAsync` or `t.notThrowsAsync` now

test.js
```diff
@@ -342,7 +343,11 @@ test('flush - respond with an error', async t => {
     }
   ]
 
-  await t.throws(client.flush(), 'Bad Request')
+  try {
+    await t.throwsAsync(() => client.flush())
+  } catch (error) {
+    console.log('Dafuq:', error)
```
Dahaden, line 349: Part of debugging in an attempt to find where the unhandled promise rejection was coming from.
This console log is not called at all. Very confused

Dahaden:
Closing this PR in favour of https://github.com/segmentio/analytics-node/pull/294

pbassut:
@Dahaden But this PR included some useful upgrades(like ava and babel), right? 
I don't think we should ditch that

Dahaden:
I can create another PR with just the ava bump if you want? The big changes are just:

1. No import (because babel got removed from ava, might be a better way to do that though?)
2. Change in API for Async functions,
3. Change in API for error matching

pbassut:
That would be terrific @Dahaden . 
We're on a really old version of Ava.

pbassut:
I like the baby steps approach here. Let's get this #288 going as it is the biggest chunk of the whole deal.

PR #294, opened 2021-09-03
#287 Flush now returns promises correctly

Dahaden:
Noticed `pify` was causing a few issues as it conflicted with the promises returned by the flush function

This change might even be worthy of a major version as before this, it was not expected for the client to reject promises returned by `client.flush`.


pbassut:
i'm thinking if tests are necessary given this function returns something totally different now.

pbassut (approved):
Looks good for now. Let's discuss if we need to add tests later.

PR #299, opened 2021-10-01
Flush was making callback to be called twice

pbassut:


nd4p90x:
PR #315 solves this, closing

Issue #309, opened 2021-11-12
`flush` does not guarantee that all inflight messages are sent


yo1dog:
`flush` does not guarantee that all inflight messages are sent before calling the given callback. Instead, `flush` simply sends a batch of queued messages and waits for only that batch's response before callback.

This is contrary to common expectations that a "flush" function completely empties the buffer to the destination. In fact, there are several open issues caused by this assumption.

Further, this means there is no way to know when both all queued and all inflight messages are full sent (fully flushed). For example, if you were to call `flush` while another flush was already in progress, the callback could be called before the previous flush was complete.

This can be a very common occurrence in short-lived environments like cloud functions.

Example:

```javascript
const analytics = new Analytics(writeKey, {flushAt: 2});
analytics.track(...); // This will trigger a flush because line 212: flushed === false
analytics.track(...);
analytics.track(...); // This will trigger a flush because line 218: queue.length >= flushAt
analytics.flush().then(() => {
  // This callback will be called immediately because line 249: queue.length === 0
  // The process would exit before any messages are sent.
  console.log('done');
  process.exit(0);
});
```

Similarly:

```javascript
const analytics = new Analytics(writeKey, {flushAt: 2});
analytics.track(...); // This will trigger a flush because line 212: flushed === false
analytics.track(...);
// The following will trigger a flush because there is a message in the queue.
analytics.flush().then(() => {
  // This calllback may be called before the other flush completes.
  // The process may exit while the message from the other flush is still inflight.
  console.log('done');
  process.exit(0);
});
// There are now 2 concurrent flushes running in parallel.
// Waiting on the final flush creates a race condition as it may finish before the previous flush.
```

I suggest adding a new function that `flush()`es all queued messages and does not callback until all concurrent flushes are complete. Also updating the documentation to make it clear that `flush` may defy assumptions about awaiting all queued and inflight messages.

felipe-najson-ntf:
Hello @yo1dog, thanks for your contribution to the library! I just tested the cases mentioned above and clearly flush creates a race condition in which 2 simultaneous downloads are executed in parallel.

I'll be working on a solution to solve this next week.

Thanks, Segment team.

benjaminhoffman:
I'm also experiencing issues with queueing and flush.  Although I'm hosted on vercel, I'm _not_ using serverless functions. Would love to see this fixed to help eliminate issues in my debugging.

yo1dog:
[...]

Issue #303, opened 2021-10-18
Events are not sent in a Vercel serverless function


andreaminieri:
Hi everybody, I'm using analytics-node 5.1.0 in a SSR application deployed on Vercel.

I wrote an API that is deployed as Vercel serverless functions.
I had some troubles to make it work and this is probably related to #245. These are the steps I followed for one function.

This is what I had initially:

```javascript
app.post('/api/test', async function (req, res) {
  analytics.track({...},
    function (err, batch) {
      if (err) {
        console.error(`Error occured in track call: ${err}`)
      }
      console.log(`Flushed from track call: ${batch}`)
    }
  )

  console.log('Sending back HTTP 200...')

  res.status(200).json({
    message: 'Signal to segment sent.',
  })
})
```
This works locally and on my server console I get

`Sending back HTTP 200...`
`Flushed from track call: undefined`
`Flushed from track call: [object Object]`

But it doesn't work on Vercel, in the serverless function log I got just

`Sending back HTTP 200...`

The `track` call is not flushed and no event is sent to Segment. This is similar to what is explained in #245, probably Vercel serverless function finishes and kills all threads before the queue is actually flushed. Following the approach in #245, I updated the function as follows:

```javascript
app.post('/api/test', async function (req, res) {
  analytics.track({...},
    function (err, batch) {
      if (err) {
        console.error(`Error occured in track call: ${err}`)
      }
      console.log(`Flushed from track call: ${batch}`)
    }
  )

  await analytics.flush(function (err, batch) {
    if (err) {
      console.error(`Error occured in flush call: ${err}`)
    }
    console.log('Flushed from flush call.')
  })

  console.log('Sending back HTTP 200...')

  res.status(200).json({
    message: 'Signal to segment sent.',
  })
})
```
First thing I noticed is that VSCode gives me a warning about using `await` on `analytics.flush`: 'await' has no effect on the type of this expression.
I'm using analytics-node 5.1.0 so flush should return a promise. Maybe this warning is due to the types definition of the function, I'm not sure whether this has an effect on the actual async behavior of the function. Anyway, this still works locally and the log is:

`Sending back HTTP 200...`
`Flushed from flush call.`
`Flushed from track call: undefined`
`Flushed from track call: [object Object]`

But it still doesn't work on Vercel, the serverless function log I got is:

`Sending back HTTP 200...`
`Flushed from flush call.`

Am I doing something wrong here?

[...]

PR #316, opened 2021-12-07
Solving race condition between flushes.

felipe-najson-ntf:
Enqueue's flush now awaits before continuing.

There are cases where race conditions can be given like the examples presented in issue #309. Implementing it asynchronously we make sure that the in-flight flushes ended before continuing.

This problem is also related to the strange behavior that the user has in issue #303

### DIFF
--- a/index.js
+++ b/index.js
@@ -79,9 +79,9 @@ class Analytics {
    * @return {Analytics}
    */
 
-  identify (message, callback) {
+  async identify (message, callback) {
     this._validate(message, 'identify')
-    this.enqueue('identify', message, callback)
+    await this.enqueue('identify', message, callback)
     return this
   }
 
@@ -93,9 +93,9 @@ class Analytics {
    * @return {Analytics}
    */
 
-  group (message, callback) {
+  async group (message, callback) {
     this._validate(message, 'group')
-    this.enqueue('group', message, callback)
+    await this.enqueue('group', message, callback)
     return this
   }
 
@@ -107,9 +107,9 @@ class Analytics {
    * @return {Analytics}
    */
 
-  track (message, callback) {
+  async track (message, callback) {
     this._validate(message, 'track')
-    this.enqueue('track', message, callback)
+    await this.enqueue('track', message, callback)
     return this
   }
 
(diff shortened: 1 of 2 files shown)


felipe-najson-ntf:
@pooyaj Please check issue #309 and confirm if this is an error or an expected behavior. Greetings

pooyaj (commented):
thanks @felipe-najson-ntf for the PR. Looks like there is a CI failure, can you please look into it? Also do we need to do a major version bump for this change?

pooyaj (approved):

yo1dog:
FYI, this pull does not solve #309 unless the expectation is that the consumer must now `await` every identify, track, etc. call. This means the user can no longer "fire and forget".  Instead, the user must now be concerned with the implantation details and potential unrelated side affects of a flush being triggered as a result of an identify/track/etc call. This also requires every calling context to be async and possibly blocking.

Namely, every call to `track` now requires the user to wait for completion of and contended with delays caused by a possible flush.

I believe this violates users' expectations. Users expect to be able to:
1. Queue messages and allow the library to transparently manage the queuing and flushing ("fire and forget").
2. Be notified when events occurred. Such as when a specific message was sent or when all messages are fully flushed.
3. Manually trigger a flush.

As such, I believe making these methods async is not the right solution. Instead, the library should track in-progress flushes and provide a method that awaits all in-progress flushes. This could be as simple as maintaining an array of all axios promises and then `await Promise.allSettled(...)`.

Further, making these methods async would cause existing code to violate ESLint's `no-floating-promises` rule (which is enabled by default via `eslint:recommended`) and would be a breaking change for many users. They would have to find every identify/track/etc call and prepend with `void` to resolve.

-------

I hope you don't take this the wrong way. Async queues can be a hard problem. Luckily they are a solved problem so there is no need to reinvent the wheel. Instead of implementing this yourself, I would use an existing solution. For example, you could use [async's cargo queue](https://caolan.github.io/async/v3/docs.html#cargoQueue) like so:

```javascript
class Analytics {
  constructor() {
    const maxConcurrentRequests = 10;
    const maxMessagesPerRequest = 100;
    
    this.cargoQueue = async.cargoQueue((messages, cb) => {
      sendMessagesToServer(messages, err => {
        // ...
      });
    }, maxConcurrentRequests, maxMessagesPerRequest);
  }
  
  enqueue(message, cb) {
    this.cargoQueue.push(message, cb);
  }
  
  flush(cb) {
    this.cargoQueue.drain(cb);
  }
}
```

Close in functionality to your current implementation. If you wanted finer control you could manually `.pause()` and `.resume()` the queue. Or use a more robust async queuing/batching library.

PR #321, opened 2022-02-08
Update code snippet to match analyics-code API

benjaminhoffman:
Following the [documentation example here](https://segment.com/docs/connections/sources/catalog/libraries/server/node/#track) as well as the official `analytics-node` API signature, this code snippet example needs to be updated. The current example is from `analytics.js`, Segment's clientside library.

NOTE: userId (or anonymousId) is required for analytics-node events


benjaminhoffman:
should we merge?

pooyaj (approved):

PR #322, opened 2022-02-09
Solving race condition between flushes V2

gongzibaimo:
Just want to fix the same issue as !316 and trying to fix the pipeline since that issue is inactive for a while.


gongzibaimo:
@nd4p90x Hello, I am trying to fix the pipeline issue from MR !316, but now I face an issue when doing `snyk auth`. I have authorized my github account with snyk but this issue is still happening, do you have any insight for this?

PR #327, opened 2022-02-23
Add noUnhandledRejection that stops throwing unhandled rejections

GCAndrew:


emont01:
Hi @GCAndrew after checking the failed CI (tests) seems like using noUnhandledRejection breaks two use cases: 

1. flush - time out if configured
2. ensure that failed requests are not retried forever

There may be some additional checks need to see if the error will trigger a retry...

nd4p90x:
Due to lack of movement on this PR, we are closing it.  If anyone determines that it needs to be moved forward, please feel free to re-open and contribute.

PR #328, opened 2022-03-02
marked queue as flushed by default and flushing all messages

emont01:
Marked queue as flushed by default and flushing all messages once 'flush()' is called


nd4p90x:
@emont01 Please review and fix the yarn errors, thank you.

Issue #326, opened 2022-02-23
Unhandled Promise Rejection Can Crash App


GCAndrew:
As described in #320 , ‘flush()’ can cause an uncaught exception. Listening for uncaught exception events isn’t an ideal error handling pattern. 

This exception should be able to be handled at the call-site of ‘track()’ and similar functions.

GCAndrew:
Example PR: https://github.com/segmentio/analytics-node/pull/327

emont01:
@GCAndrew your PR adds a property that could be used to "turn off" error throwing once the axios request (promise) is rejected, I got nothing against this idea.

Reading the code, I am getting the impression that the idea here is to pass the possible error to any callback sent to flush and the individual message's callbacks, then throwing the exception again allows the promise to continue its normal course, which seems to be the problem here; my only question is: wouldn't it be better just to add an optional parameter to flush instead of making this a concern of the analytics client.

```js
  flush (callback, noUnhandledRejection = false) {
    callback = callback || noop
    ....
    .catch(err => {
        if (err.response) {
          const error = new Error(err.response.statusText)
          done(error)
          if (!noUnhandledRejection) {
            throw error
          }
        }

        done(err)
        if (!noUnhandledRejection) {
          throw err
        }
    })
```

emont01:
@GCAndrew sounds good to me! I would suggest your PR makes the behavior the default 
```js
this.noUnhandledRejection = options.noUnhandledRejection || true
```

GCAndrew:
@emont01 Great! Update my PR: https://github.com/segmentio/analytics-node/pull/327/commits/d6b7810b048153600d973a7025dfd86c222a460b

emont01:
@GCAndrew yes!  The default value of noUnhandledRejection should be false IMHO, my main concerns right now is how setting it to true will break the retry behavior, we should let axios-retry do its thing before  handling any rejection.

To be honest I completely forgot about 'axios-retry' being used and after going over how that works [1] now I think some changes maybe necessary in either the tests, your PR or both.

axios-retry documentation does not mention any way to capture the "final error",  at that moment we could handle the exception and avoid an unhandled rejection, possibly something like [2].

I will try with my suggested try-catch changes and let you know if they work, let me know if you see a better way.

[1]
1. axios-retry catches any error and then checks if it is possible to retry
2. axios-retry will then return a promise with a delay or reject

[2]
``` js
  flush (callback) {
    ....
    try {
[...]

Issue #320, opened 2022-01-20
analytics-node might cause an unhandledRejection


yigaldviri:
Hi there, 

Im trying to use this package to report events to Segment and as a consumer I'm trying to protect my server from errors. My code looks like this:

```
track = promisify(analytics.track.bind(analytics));

async report (userId: string, eventType: SegmentEvent, properties) {
    try {
      await track({
        userId,
        event: eventType,
        properties: addLinkToProperties(eventType, properties),
      });
    }
    catch (e) {
      log.warn('Couldn\'t report to segment about event', { eventType, properties, e });
    }
  }
```

I had a look at the package's code and I think that In some cases the package can cause an unhandledRejection error. For examplee, in case of a downtime on Segment servers.

I simulated it by changing the code and throwing an error inside the `then` clause of the `flush` function:
![image](https://user-images.githubusercontent.com/13430395/150319930-e3c849cf-f40c-403d-9a25-13701f88188b.png)

Tested it an indeed I got `[ERROR] - unhandledRejection: Error: test an error`

This kind of error might result in a server crush and using `try/catch` wont help either so am I missing here something? Is there a way to avoid that error? Or maybe my case can't really happen in reality?

nd4p90x:
Please re-open if more discussion is needed.  Thank you

GarrettCox:
What is the solution here? Apps can have intermittent network disruptions, and a fatal unhandled promise rejection is a poor user experience.

supermacro:
> Please re-open if more discussion is needed. Thank you

As requested by @nd4p90x ... looks like more discussion is needed. Could we re-open?

edsonjab:
Hi @GCAndrew , @supermacro , @yigaldviri ,  I found a solution for this issue. You can handle the error with the next code:

![image](https://user-images.githubusercontent.com/22898327/171959890-013034e2-2fb4-4402-bd9a-0c38e03e6a63.png)

The node process global has [an unhandledRejection event](https://nodejs.org/api/process.html#process_event_unhandledrejection) for unhandled promise rejection.

tam-carre:
@edsonjab it works but as said in #326, 

> Listening for uncaught exception events isn’t an ideal error handling pattern.

It is very user-unfriendly to require that the library's exceptions be handled inside the `unhandledRejection` handler.

rdsedmundo:
[...]

PR #342, opened 2022-06-17
Add an errorHandler property to constructor options

tam-carre:
✔️ All existing tests pass
✔️ New functionality is unit-tested

Addresses #320 #326 

# What this adds

This PR adds an optional `errorHandler` property to the class constructor's options.
If unspecified, the behaviour of the library does not change.
If specified, when an axios request fails, `errorHandler(axiosError)` will be called instead of re-throwing the axios error.

Example usage:

```js
const Analytics = require('analytics-node');

const client = new Analytics('write key', {
  errorHandler: (err) => {
    console.error('analytics-node flush failed.')
    console.error(err)
  }
});

client.track({
  event: 'event name',
  userId: 'user id'
});
// ☝️ if this fails when flushed no exception will be thrown and instead
// the axios error will be logged to the console.

```

# Why

Since no way to handle the re-thrown axios error was exposed previously, this meant that if the axios request failed for any reason, the application using analytics-node would crash. A workaround was to use node's process `unhandledRejection` or `unhandledException` handlers, but this is very un-ergonomic for library users.

# Work remaining

This functionality needs to be documented: https://segment.com/docs/connections/sources/catalog/libraries/server/node/

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,7 @@ class Analytics {
    *   @property {Object} [axiosInstance] (default: axios.create(options.axiosConfig))
    *   @property {Object} [axiosRetryConfig] (optional)
    *   @property {Number} [retryCount] (default: 3)
+   *   @property {Function} [errorHandler] (optional)
    */
 
   constructor (writeKey, options) {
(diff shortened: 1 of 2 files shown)


pooyaj (approved):
Looks great 🙌

nd4p90x:
PR For Node Documentation Update:
https://github.com/segmentio/segment-docs/pull/3111

PR #347, opened 2022-07-27
Ensure callback is called when invoking errorHanlder

timhaley94:
I work at Trello @ Atlassian.

We use this library, however, it started causing problems for us when `throw err` was added to the `.flush()` method. This has been documented here. (#320) For us it was resulting in process crashes when we receive an error message (like a 429) when emitting an event. We had to pin the version of this library to a version before this behavior was added.

I saw, that the `errorHandler` property was added (#342) and was attempting to update our code to use it, however, when writing a test case, I saw that my event callback was never invoked when `errorHandler` was present and called 

This is because, in the `.flush()`, everywhere else the method can end, we call `done(err)`, however we do not when looking for and invoking the `errorHandler`. To reiterate the importance, it's a core bit of functionality of this library that the callbacks for events are invoked when the messages are flushed, so it's needs to occur in this instance as well.

### DIFF
--- a/index.js
+++ b/index.js
@@ -298,6 +298,7 @@ class Analytics {
       })
       .catch(err => {
         if (typeof this.errorHandler === 'function') {
+          done(err)
           return this.errorHandler(err)
         }
 
--- a/test.js
+++ b/test.js
@@ -381,6 +381,23 @@ test('flush - do not throw on axios failure if errorHandler option is specified'
   t.true(errorHandler.calledOnce)
 })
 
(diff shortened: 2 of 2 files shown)


timhaley94:
I'm not sure if it's important that I open an issue before a PR. If it is, I'm happy to retroactively open an issue.

pooyaj (approved):

pooyaj:
thanks for the fix 🙌

PR #350, opened 2022-10-03
minimize number of calls to JSON.stringify

schmod:
Small performance tweak.

Instead of calling `JSON.stringify()` on the entire queue of messages every time we enqueue a message, analytics-node will now keep a running tally of the queue size, and only call `JSON.stringify()` once when the message is enqueued.

This should offer a minor performance-improvement for anybody enqueuing lots of messages.  

---

As an aside, I'm also wondering if this bit of code is actually relevant any more:

```js
    if (!message.messageId) {
      // We md5 the messaage to add more randomness. This is primarily meant
      // for use in the browser where the uuid package falls back to Math.random()
      // which is not a great source of randomness.
      // Borrowed from analytics.js (https://github.com/segment-integrations/analytics.js-integration-segmentio/blob/a20d2a2d222aeb3ab2a8c7e72280f1df2618440e/lib/index.js#L255-L256).
      message.messageId = `node-${md5(JSON.stringify(message))}-${uuid()}`
    }
```

Given that nobody should be running this library in a browser (or on a server without a good source of entropy), can't we just use a plain `uuid()` here?  

I'm afraid to touch this, given that it'd be an actual functionality change, but I think you could probably shave some bytes off of your payloads, and squeeze out a few more CPU cycles if you got rid of the stringify/hashing here!

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,7 +38,7 @@ class Analytics {
     assert(writeKey, 'You must pass your Segment project\'s write key.')
 
     this.queue = []
-    this.queueBytes = 0;
+    this.queueBytes = 0
     this.writeKey = writeKey
     this.host = removeSlash(options.host || 'https://api.segment.io')
     this.path = removeSlash(options.path || '/v1/batch')
@@ -218,14 +218,13 @@ class Analytics {
       return
     }
 
-    
     const hasReachedFlushAt = this.queue.length >= this.flushAt
     if (hasReachedFlushAt) {
       this.flush()
       return
     }
-    
-    this.queueBytes += JSON.stringify(item)
+
+    this.queueBytes += JSON.stringify(message)
     const hasReachedQueueSize = this.queueBytes >= this.maxQueueSize
 
     if (hasReachedQueueSize) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -254,7 +262,8 @@ class Analytics {
       return Promise.resolve()
     }
 
-    const items = this.queue.splice(0, this.flushAt)
+    this.queueBytes = 0
+    const items = this.queue.splice(0, this.queue.length)
```
schmod, line 266: I changed this away from `this.flushAt`, because it _should_ be functionally-equivalent under the current implementation, and also because it could potentially lead to confusing bugs where calls to `flush()` might actually leave items on the queue (which seems like it would _never_ be desirable)

edsonjab (approved):
All works!!

nd4p90x:
@schmod Do you mind resolving the conflicts so we can merge in this PR?  Thanks.

schmod:
Done! Sorry for the delay!

schmod:
Bump!

silesky:
@schmod, I recommend that you migrate to @segment/analytics-node; it has about 20% less overhead. This package will be deprecated shortly.

schmod:
Neat!  We'll consider migrating once the new package is out of beta.

Issue #349, opened 2022-08-18
Typescript support


agyimr:
We just had a major issue in production, causing some of our customers to not be able to access our application, due to a faulty segment tracking call, which failed due to the missing userId field. The blame is of course partially ours, since we should have implemented all the tracking in a fail-safe way, but I can't shake the feeling that should the package have correctly typed interfaces, we could not even commit faulty tracking code into the app. 

We are using the @types/analytics-node package, in which all userId fields are optional.

Could you make sure that the interfaces reflect what you expect as tracking calls?

edsonjab:
Hi @agyimr  thanks for your report, we start looking into this

Jackman3005:
Typescript support is incredibly minimal. It would be really nice if we had types for things that are already well documented. Like [traits](https://segment.com/docs/connections/spec/identify/#traits) and [context](https://segment.com/docs/connections/spec/common/#context) for starters.

Is there a reason why the types are not included in the library by default? This has become commonplace for most modern libraries and even benefits Javascript code bases since most IDEs will still use type information for context tips/autocomplete. When the types and js code are in separate libraries it adds the hassle of ensuring that the type library version is for the same version of code in the Javascript library.

We're trying out Segment and having to switch clients from Amplitude to Segment and the developer experience in this space is much better from Amplitude.

yujidude:
The TypeScript definition in the DefinitelyTyped project is either currently (or originally) for analytics-node 3.1 which may have resulted in some of your issues @agyimr. Given that the project is already following JSDoc for documentation it might make sense to generate .d.ts files directly to avoid maintaining multiple sets of documentation.

More information on generation can be found here: https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html

silesky:
@agyimr Thank you for your patience. If you or anyone in in this issue is interested, I'd like to invite you guys to switch to the new [analytics.js typescript SDK](https://github.com/segmentio/analytics-next/tree/master/packages/node), written 100% in typescript. If you run issues / have feedback, leave comments in the [analytics-next](https://github.com/segmentio/analytics-next) repo!

Jackman3005:
@silesky thanks for getting back to us. Will take a look at the PR and the new TS library. 

Is the new library where most of the energy is going for Segment? I've noticed some other issues getting lackluster response in this repo, hopefully the next iteration brings nice improvements.

Cheers,
Jack

silesky:
[...]

PR #356, opened 2023-01-12
Add typescript support

spencerattick:
In response to a [couple of folks asking for it](https://github.com/segmentio/analytics-node/issues/349), I've added Typescript support to analytics-node. I used [this documentation from Typescript](https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html) as a guide. 

Nearly all errors have been resolved in terms of Typescript's requirements and analytics-node's test suite. A few exceptions were made to Typescript's requirements in which cases the errors were handled by disabling some Typescript checks at the top of index.ts. 

I recently saw that we're moving away from analytics-node to the new SDK so didn't perfect this new version. If there is interest, I'm happy to keep working on this, otherwise I'll just leave things here.


silesky:
Thanks @spencerattick. I wouldn't put any more work into this =)

